PR #268, opened 2019-12-08
Open-source more prominent.

beijbom:
Reorganized about page a bit and made the open-source stuff more prominent.


StephenChan:
I like most of these changes. I'm not entirely sure about the prominent GitHub icon near the top of our landing page, though. Maybe it's just me, but it's almost suggesting that we're affiliated with GitHub or something. I'd say either move it to a less prominent place, or just remove it.

If we remove it, the GitHub repo is roughly as easy to find as our Forum, and slightly less easy to find than the Blog and About sections. That seems OK to me.

StephenChan:
I like the About edits, though. Moving the site features up near the top is good. And I noticed the double 'repository' in the first paragraph but forgot to mention it.

beijbom:
@StephenChan : makes sense. I removed the icon on the front page. And did some cosmetics on the about page.

StephenChan:
- 'A web solution' is correct: https://www.merriam-webster.com/words-at-play/is-it-a-or-an (like the 'a one hundred dollar bill' example)
- I was OK with the 'open source' link on About, for the record, but either way works.

beijbom:
> * 'A web solution' is correct: https://www.merriam-webster.com/words-at-play/is-it-a-or-an (like the 'a one hundred dollar bill' example)
My bad. Typo. Fixed.
> * I was OK with the 'open source' link on About, for the record, but either way works.
Ok. Yeah, I thought it looked odd since there was only one link.

PR #270, opened 2019-12-08
Continuous Integration with Travis

beijbom:
This PR adds CI to the repo. I added a new config file `travis.py` and had to do some workarounds for `secrets.py` in `settings/base.py`. But it feels pretty clean to me, so I think it is ok to keep this way. 

I'm building with `pip install -r requirements/production.txt` which seems to work fine.

As discussed below there are two tests that fail currently. But it's not related to this PR, so I think we can merge.


StephenChan:
Ah, if you're getting test failures on 2 Unicode related tests, I believe the staging server gets those 2 failures as well. Haven't figured out why that happens.

Anyway, good call, Travis seems like something we should be using. I just thought it'd be tricky to set up, but this looks like good progress!

beijbom:
@StephenChan : yeah, it was actually pretty easy. Travis is free for open-source projects, so I thought I'd try it. You can go to https://travis-ci.org/beijbom/coralnet and see the builds. 
Right now it all compiles and builds, but one weird thing left. Those 2 tests fail, plus an extra dummy-test I added, but the build passes. Hmm.

I'll keep looking at it, but holler if you figure it out.

```
FAIL: test_will_fail (lib.tests.tests.FailingTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/beijbom/coralnet/project/lib/tests/tests.py", line 26, in test_will_fail
    self.assertEqual(True, False)
AssertionError: True != False
----------------------------------------------------------------------
Ran 681 tests in 77.964s
FAILED (failures=3, skipped=14)
Destroying test database for alias 'default'...
The command "python manage.py test" exited with 0.
Done. Your build exited with 0.
```

According to the django docs manage.py test should exit with 1 of one or more tests fail:
https://docs.djangoproject.com/en/3.0/topics/testing/overview/
_"Note that the return code for the test-runner script is 1 for any number of failed and erroneous tests. If all the tests pass, the return code is 0. This feature is useful if you’re using the test-runner script in a shell script and need to test for success or failure at that level."_

StephenChan:
> Right now it all compiles and builds, but one weird thing left. Those 2 tests fail, plus an extra dummy-test I added, but the build passes. Hmm.

I can't seem to find any satisfying solutions for this, but there are a couple of possible workaround avenues here: https://stackoverflow.com/questions/56208271/cucumber-tests-fail-but-travis-build-still-passes

beijbom:
@StephenChan : I think I figured it out. We have inherited 
`class TempStorageTestRunner(DiscoverRunner):`
in `lib/tests/utils.py` and forgot to capture the return in this call:
```
super(TempStorageTestRunner, self).run_tests(
                test_labels, extra_tests=extra_tests, **kwargs)
```
It works now.

Now we just need to actually fix that unit-test. haha. Since it pass on our dev. machines it's guaranteed to be a linux thing.

StephenChan:
Oooh, right, good call. I didn't think of that.

PR #271, opened 2019-12-08
Changed character order in text fixtures such that tests pass in Travis.

beijbom:
Seems the problem is related to how the DB or OS sort unicode chars with respect to "normal" chars. This PR changes the text fixtures to accommodate the build environment. More work is needed to fix across environments.

### DIFF
--- a/project/export/tests/test_covers.py
+++ b/project/export/tests/test_covers.py
@@ -157,7 +157,7 @@ class UnicodeTest(BaseExportTest):
         response = self.export_image_covers(post_data)
 
         expected_lines = [
-            'Name,Annotation status,Annotation area,A,い',
+            'Name,Annotation status,Annotation area,い,A',
             'あ.jpg,Confirmed,X: 0 - 100% / Y: 0 - 100%,60.000,40.000',
         ]
         self.assert_csv_content_equal(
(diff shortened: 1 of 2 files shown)


beijbom:
@StephenChan : switching the order of the expected results helped the tests pass on travis. It seems you are sorting by short-code. So the default order there seem to be OS dependent. Can we re-write the test to not expect a particular order (since it doesn't really matter, or does it?) ?

beijbom:
@StephenChan : indeed, the second one is for the same reason. Mac, PC orders the labels `[A,い]`, while linux does `[い,A]`.

Possibly related: https://stackoverflow.com/questions/22534484/postgresql-incorrect-sorting

beijbom:
@StephenChan : this is not really a fix since the tests now fail on the dev-machines. But I prefer that the CI passes.

StephenChan:
Oh, that's really interesting. IMO, it seems best to treat this as a bug, and make sure all of our systems (dev and production) sort in a consistent way. Even if these tests don't highlight cases where the sort order is very important.

But yeah, this is fine for now. I can take a look at this sorting business sometime next week.

beijbom:
Sounds good. I'll merge then. Can you create a ticket for the bug? We may want to wait until we update to python3 and the more recent Django. I'm not sure if it is related, but it may be something that was fixed as part of regular updates.

StephenChan:
Created issue #272. Good point about Python 3; we'll want to upgrade to that soon as well. I suppose I'll take a quick look at this bug first, and if no luck, I'll put it off until Py3.

Issue #275, opened 2019-12-14
Image patch creation issues for RGBA images


beijbom:
As the title suggests, the patch generation fails for RGBA images. 

The error triggers on this page: https://coralnet.ucsd.edu/label/3679/example_patches_ajax/?page=1

Here's the log for one of the RGBA errors: https://coralnet.ucsd.edu/admin/errorlogs/errorlog/2250/change/

If you expand the local vars under `generate_patch_if_doesnt_exist(point.pk)`, you should see the image and source in question.

beijbom:
@StephenChan : what is our routine for text fixtures? In this case, it would be nice to store the problematic image itself somewhere and write a unit-test that makes sure it can be cropped without problems. But I don't see any such mechanisms in the repo. Do you have any thoughts on this? The very simplest way would be to create a `fixture` subfolder in the `test` folders, and simply commit the image to the repo. But there may be better ways.

beijbom:
Ok. So I figured this one out. It has to do with the Pillow package upgrade. Refer to this thread:
https://github.com/python-pillow/Pillow/issues/2609

Copying the image locally, and using an older version of PIL I was able to crop and save fine. When I updated the reqs. to the current version I got the same error as in production. The fix is simple:
```
im = im.convert('RGB')
```
I will push a fix soon.

StephenChan:
> what is our routine for text fixtures? In this case, it would be nice to store the problematic image itself somewhere and write a unit-test that makes sure it can be cropped without problems. But I don't see any such mechanisms in the repo. Do you have any thoughts on this? The very simplest way would be to create a `fixture` subfolder in the `test` folders, and simply commit the image to the repo. But there may be better ways.

We don't currently have a mechanism for this. Our image format tests are currently pretty limited, only consisting of images that are trivial to create programmatically, which is why we haven't needed such a thing yet. (See `upload/tests/test_images.py` - tests any .jpg, any .png, any .bmp, image of specific resolution, empty file.)

For this issue's purposes, I think we can also create the image programmatically during the test - seems like it just has to define RGBA data, nothing too fancy. Though I do like your idea in general. We might run into more subtle image decoding challenges later (especially if we get into something like EXIF support), and sometimes it's nice if you can manually inspect what's being tested. I'm OK with a `fixtures` subfolder in the `tests` folders, since I haven't heard of some other specific way to set this up. I'll leave it up to you on whether to add such a thing now or later.

[...]

PR #276, opened 2020-01-10
crop patch bugfix and cleanup

beijbom:
Fixes https://github.com/beijbom/coralnet/issues/275.

In addition I made a few changes:
- Used snake_case for all variables
- Moved the constants to settings/base.py
- Simplified logic to crop at the center pixel. This means there may be a black area of pixel is on the edge. But I think that is better since the annotated substrate will always be at the center. (See screenshot below)
- Cleaned up code to use less local vars. Mostly a preference thing, but I don't like assigning local variables if they are only used once. E.g. `x = point.column`.

I tried this in my dev. environment and it appears to do the right thing. 

What I don't like about this PR is that there is no new tests. I was considering moving all image manipulation out to a separate method and test those, but I could not find a meaningful way to do so since at this point there are only four calls to PIL. (open, convert, crop, save).

Screen Shot 2020-01-10 at 2 36 41 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@ docs/_templates
 # Compiled Python
 *.pyc
 
+# Jupyter
+*.ipynb
+
 # Redis dumps
 *.rdb
 
--- a/project/config/settings/base.py
+++ b/project/config/settings/base.py
@@ -587,3 +587,8 @@ LIKELY_TEST_SOURCE_NAMES = ['test', 'sandbox', 'dummy', 'tmp', 'temp', 'check']
 
 # NewsItem categories used in NewsItem app.
 NEWS_ITEM_CATEGORIES = ['ml', 'source', 'image', 'annotation', 'account']
+
+# Label patch settings
+LABELPATCH_NCOLS = 150  # Size of patch (after scaling)
+LABELPATCH_NROWS = 150  # Size of patch (after scaling)
+LABELPATCH_SIZE_FRACTION = 0.2  # Patch covers this proportion of the original image's greater dimension
--- a/project/images/views.py
+++ b/project/images/views.py
@@ -160,9 +160,11 @@ def source_main(request, source_id):
 
     # Images' annotation status
     browse_url_base = reverse('browse_images', args=[source.id])
+
     def image_count_by_status(annotation_status):
         return image_search_kwargs_to_queryset(
             dict(annotation_status=annotation_status), source).count()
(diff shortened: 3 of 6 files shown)


StephenChan:
Sorry for the delay here. But yeah, thanks for looking into this. Looking forward to reducing those error emails.

I see - so it's entirely possible we had RGBA images before, but commit 8c6c7b844d9792f01050c18fdaba74f2abd33da4 (2 months ago) updated Pillow from 4.1.1 to 4.3.0, thus exposing this problem.

Converting to RGB sounds good. I can't think of a need for patch images to support transparency. It might be nice to follow [this comment's advice](https://github.com/python-pillow/Pillow/issues/2609#issuecomment-313922483) about filling transparent pixels with a fixed color (like black), but I don't think CoralNet gets a lot of images with transparent pixels, so I'm fine either way for now.

About cropping at the center pixel: Yeah, I'm fine with this change. I was probably on the fence about how this should work when it was implemented originally. But it seems better to be clear and consistent (patch center = point of interest) even if it means sacrificing a bit of aesthetic.

Most other cleanups look good.

My suggestions / things to point out:

- I think you forgot about the PEP 8 line length (79 characters).

- It looks like we didn't have tests specifically on patch generation to begin with. We just had some tests that require patch files to be generated, but we didn't check the patches themselves. I'm OK with putting off the tests till later if you're tight on time at the moment, since it'd be nice to reduce the error emails. But if you're up for it:

  - The `upload/tests/test_images.py` file has some examples of creating images in a unit test. The fixtures idea you brought up in issue #275 can also work.

  - To test cropping, we could use the `mock` module to assert that `im.crop()` is called with the expected arguments. I have used the `mock` module for one test in `test_annotation_tool.py`, but that use case isn't all that similar to the use case I'm suggesting here. In any case, our project has the `mock` package already. Although mock is a third-party package in Python 2, it's supposed to work the same as what's described in the [Python 3 docs](https://docs.python.org/3/library/unittest.mock.html).

beijbom:
Thanks @StephenChan . I didn't so much forget about the line-length as ignored it. haha. I'm just having a hard time getting used to it. But I'm getting there -- I pushed a fix which stays within 79 chars. 
I also created some tests that will catch the problem that occurred. Please do take a new look.

StephenChan:
The new changes look great - we've even got grayscale support now! And the tests run fine on my Windows machine.

Issue #231, opened 2019-08-15
API endpoint to deploy classifier


StephenChan:
Currently, machine-classifying an image using CoralNet requires uploading that image to a source. The image can then be classified using that source's classifier, and the image is also used to further train the source's classifier.

With this API endpoint, you would specify a source's classifier to use, upload an image to be classified, and get a classification result back. The image to be classified doesn't get saved to any source. Benefits:

- The set of images used for training can be different from the set of images one wants to classify.
- Not every image one wants to classify has to be uploaded permanently to a CoralNet source, nor does it have to go through the whole training pipeline. This reduces CoralNet's storage costs and training-related compute costs, and speeds up the workflow.
- Since uploads are done via API, the image upload process can be made more robust using a script with error handling, as opposed to a web-form upload which may have to be supervised manually.

The majority of issue #189 (API basics) would have to be done before this issue.

beijbom:
@StephenChan : is this issue complete?

StephenChan:
Looks like it. This issue doesn't really get into specific requirements, and deploy is indeed working.

Issue #189, opened 2018-06-23
Start a public API with authentication and rate limiting


StephenChan:
The goal of this issue is to get at least one API endpoint fully functional, and lay the groundwork for implementing more API endpoints. Rough steps to take:

- [x] Start using the [Django REST Framework](https://www.django-rest-framework.org/).
- [x] Implement any endpoint without authentication for a start.
- [x] Support session authentication for API endpoints, so that the CoralNet website can easily use its own API endpoints.
- [x] Support token authentication for API endpoints, so that a non-web app such as a command line script can securely access the endpoints.
- [x] Require authentication on existing endpoints. For example, a user should only be able to deploy a classifier if they're able to view that classifier's source.
- [x] Impose API usage-rate limits. This prevents server overload in case a buggy or malicious API client makes too many API calls. Refer to this section for implementation: http://www.django-rest-framework.org/api-guide/throttling/#throttling

Response formats and codes should be consistent and mostly be guided by some kind of standard, such as the [JSON API spec](https://jsonapi.org/). This spec doesn't have official capacity, but it's the most popular open spec for web APIs, as far as I know. IMO, we can deviate from this if there's a decent reason to do so, but all else being equal we should try to stick to this spec.

Other implementation ideas, but not necessarily required right away:

- Implement a 'services' layer as outlined here for example: https://stackoverflow.com/questions/30259452/proper-way-to-consume-data-from-restful-api-in-django
- [`django-rest-framework-json-api`](https://github.com/django-json-api/django-rest-framework-json-api) might make it easier to adhere to the JSON API spec.

Longer-term goals to consider after this issue:

- Allow other web apps like Squidle+ to programmatically work with CoralNet's automated classification system. This'll require OAuth2 (#267), and enough API endpoints (related to sources, classifiers, etc.) for such a web app to effectively 'browse' CoralNet via the API.

- Modularity: Separate CoralNet into two entities, one with the entire API and database, and one with the entire web app frontend.

  - Suppose we're slowly porting `labels` views to API endpoints. We could create a `labels_frontend` app and start moving frontend views there. `labels` remains as the backend/API so that models don't have to move to another app. Then we'd change frontend-view implementations such that they use the API endpoints instead of using the Django ORM directly. We'd also have to move utility functions and tests between apps as appropriate.

StephenChan:
Narrowed the scope of this issue a bit more (see #277).

[...]

PR #278, opened 2020-01-15
API: basics, token authentication, basic rate limiting, and start on deploy

StephenChan:
For #189, and a start on #231. If you're wondering why I worked on both at once: I needed an endpoint to test the base API functionality, and I figured it might as well be the deploy endpoint.

Update notes:

- The package `djangorestframework` was added (for various API conveniences).

- The package `six` was updated (for the collections_abc import).

- A new migration was added to the new app `api_core` (the new models represent asynchronous jobs whose statuses can be API-queried).

- Unit tests are now PostgreSQL-only (cannot run using SQLite) for the time being. I used JSONField in the new models, and that field is only supported in PostgreSQL with our current setup. When we are on Python 3, there should be a couple of solutions available to get JSONField working in SQLite. (I will make a separate issue with the details on this.)

Summary of changes:

- Implements a REST API with two types of authentication, 'session' and 'token'.

  - In session authentication, you simply log into the website as normal using a browser, then navigate to an API URL.

  - In token authentication, you make a POST request to the `token_auth` view with your username and password, and you get a token in the response. You then include this token in an `HTTP_AUTHORIZATION` header for any subsequent requests. This is better suited towards external applications such as command line, because this way they don't have to deal with cookies or CSRF protocol.

  - For simplicity, no OAuth2 for now, as discussed in [this comment](https://github.com/beijbom/coralnet/issues/189#issuecomment-562326884). Essentially, OAuth2 is tokens + third-party app authorization.
  
- Creates an ApiJob model to store job status and results, and an ApiJobUnit model to store status/results of a job's individual tasks (e.g. extract features for image 1, classify image 2, etc.). Jobs and job units are cleaned up (via `clean_up_old_api_jobs`) 1 month after the job has finished.

  - An alternative design would be to have only one model, ApiJob (without ApiJobUnit), and have all of an ApiJob's async tasks update fields on the ApiJob. However, this would have required a lot more care to prevent race conditions.

- Creates deploy, deploy_statuses, and deploy_results API endpoints according to our API spec draft. (I've made a few minor changes/suggestions to the spec, by the way.)
  
  - Implements permissions as follows: you need to be able to view a classifier's source to deploy that classifier, and you need to be the user who requested a deploy to view its status/results. Also, you need to be authenticated (via token or session) to use any of the deploy endpoints.
  
- Implements basic rate limiting: a 'burst' rate of 50 requests per minute, and a 'sustained' rate of 500 requests per hour. We can tweak the numbers later, of course.

Parts of deploy (#231) that aren't done yet:

- Actually downloading images from the URLs provided in the deploy request.

  - As an aside: Depending on what our users want exactly from a deploy function, we might consider making another version of deploy which takes CoralNet image IDs instead of URLs.

- Integrating with the `vision_backend` app to call spacer (or call a mock backend).

  - Relatedly, a note about the separate `vision_backend_api` app I created: In the future, if we commit more to an API-driven project design, we may want to make `vision_backend` hold the models, utils, and API views, while a new app called `vision_backend_web` holds the HTML-served views. This way the models don't have to migrate their app name. However, we're just getting started with the API now, so this separate VB-api app seems more appropriate for the time being.


StephenChan:
I screwed up the `clean_up_old_api_jobs` task's logic. Need to fix that real quick.

StephenChan:
Should be fixed.

beijbom:
@StephenChan : I'll try to figure it out, but can you add a "how to use" somewhere? Either in this PR description, a `README.md` in the app, or add a script in the `scripts` folder?

beijbom:
Figured out the first part:
`curl -d "username=[username]&password=[password]" http://127.0.0.1:8000/api/token-auth/`
I should be able to figure out the rest from here.

StephenChan:
@beijbom  Think I still need your review on this before it can be merged.

beijbom:
@StephenChan : done, you can go ahead and merge. nice work.

PR #279, opened 2020-01-19
Deploy call

beijbom:
This PR closes the loop on the deploy call. It should be reviewed together with https://github.com/beijbom/ecs_spacer/pull/1 and https://github.com/beijbom/ecs_spacer/pull/2.

Main features:
- A new deploy task is created in spacer. This extracts features and does the classification in one go.
- A corresponding deploy task is created in `coralnet.vision_backend`. Also added support for reading the results from the spacer job and updating the DB on the CoralNet side.
- Support is added for failed jobs. See example below.
- The `vision_backend_api` app is updated to use the new deploy call. All tests are updated. Some tests are skipped since we don't have a mock or offline version available of the deploy call (yet).
- All tests in the `vision_backend` app are moved to subfolder and cleaned up to conform with PEP8.
- Two new tests written, mostly to test the parsing of the spacer results jobs.

How to test:
- Run the dev server with the s3 backend, make sure celery and celery beat are running.
- Submit API calls to localhost (or http://127.0.0.1:8000/).

Future work:
- Along the lines of previous discussions, we need to make sure that the spacer backend can be run without going all the way to s3/AWS. I'll work on that next.

Failed job example:
```
curl -v -H "Authorization: Token xyz" -d 'images=[{"url": "https://coralnet-beijbom-dev.s3-us-west-2.amazonaws.com/media/images/nothing_here.jpg","points": [{"row":100,"column":100}, {"row":200,"column":200}]}]' http://127.0.0.1:8000/api/classifiers/14/deploy/
```
follows by the corresponding `results` call gives:
```
{"data":{"images":[{"url":"https://coralnet-beijbom-dev.s3-us-west-2.amazonaws.com/media/images/nothing_here.jpg","error":"IOError(\"cannot identify image file u'nothing_here.jpg'\",)"}]}}
```

=== original note ===
@StephenChan : I'm opening this PR now for visibility. Together with the change in https://github.com/beijbom/ecs_spacer/pull/1 this closes the loop such that deploy calls can be sent to spacer and received after completion. Remaining to do some shakeouts & tests as well as changing the outdated logic in the status and results view.


StephenChan:
So far so good. I'm sure you've got other cleanups in mind, but I'll just point out one thing I noticed: in `vision_backend_api/views.py`, `features_job_unit` should be renamed to something else. `deploy_job_unit`, or just `job_unit` or `unit` since we already know it's for deploy, etc.

project/vision_backend_api/tests/test_deploy.py
```diff
@@ -354,7 +355,8 @@ def test_deploy_response(self):
             reverse('api:deploy_status', args=[deploy_job.pk]),
             "Response should contain status endpoint URL")
 
-    @patch('vision_backend_api.views.deploy_extract_features.run', noop_task)
+    @skip("Deprecated since deploy now a single call.")
```
beijbom: @StephenChan : Did I understand this correctly that this is deprecated due to single call? If so, should I just delete it?

project/vision_backend_api/tests/test_deploy.py
```diff
@@ -400,7 +402,8 @@ def test_pre_extract(self):
                 image_order=0),
             "Unit's request_json should be correct")
 
-    @patch('vision_backend_api.tasks.deploy_classify.run', noop_task)
+    @skip("Deprecated since deploy now a single call.")
```
beijbom: @StephenChan :  Did I understand this correctly that this is deprecated due to single call? If so, should I just delete it?

StephenChan: I think `test_pre_extract` and `test_pre_classify` should be replaced with a `test_pre_deploy` test. It should be quite similar to `test_pre_extract`: check that the ApiJob was created with correct field values, and check that the ApiJobUnit was created with correct field values (the unit type should now be deploy instead of feature extract).

beijbom: Fixed.

StephenChan: Looks good, thanks.

beijbom:
@StephenChan : this is ready to review.

StephenChan:
Nice work! Everything checks out from what I'm seeing. I'm too lazy to call spacer myself to double-check that part, but the rest of the tests all work for me.

Issue #289, opened 2020-02-15
Can't remove label from labelset if it still has unconfirmed annotations


StephenChan:
I tried out PR #285 (management command to replace annotations) in production, and it works, but it made me think of something else.

The management command only changes CONFIRMED annotations for a label. I did it this way because we probably want unconfirmed annotations to be synced with the Score and valresult structures, and the easiest way to make sure those things are all synced is to just train and accept a new classifier.

One problem, however: currently, the labelset 'Add/Remove Labels' page only lets you remove a label from the labelset if it has no annotations in the source whatsoever. This includes unconfirmed annotations. Also, the label might not have enough impact on overall classifier accuracy to allow a new classifier to get accepted. So there's no guarantee that the unconfirmed annotations would get changed anytime soon.

It seems we should modify the page to let users remove a label if it has no **confirmed** annotations - unconfirmed would be OK.

Once the user submits their change to the labelset, the backend data for the source will get reset, including erasing the unconfirmed annotations for that unwanted label... after an `apply_async()` call, that is. There may be a brief period where the source has unconfirmed annotations for a label that's not in the labelset. I'm not sure whether that could lead to any strange errors. If so, we could possibly delete the unconfirmed annotations for that label synchronously, as part of the labelset change request, instead of doing it asynchronously.

@beijbom Does this all make sense? Does anything else come to mind that we should watch out for here?

beijbom:
Great points, thanks for bringing this up. Two comments. 
- Ideally it should be enough to switch all CONFIRMED AND UNCONFIRMED annotations for the source to the new label. But since there are some meta-data stored outside the DB schema (e.g https://github.com/beijbom/coralnet/issues/288, but also the internal representation of the classifier objects from scikit learn), we need to err on the side of wiping out all classifiers and starting over, just like we do with any label-set change. In other words, we need to call https://github.com/beijbom/coralnet/blob/master/project/vision_backend/tasks.py#L321 as part of that management command. This deletes all robot annotations for that source, so you will end up deleting all CONFIRMED annotations of the old label and ALL UNCONFIRMED annotations for the whole source. I'm sorry I didn't catch this during review.
- Regarding your comment on "It seems we should modify the page to let users remove a label if it has no confirmed annotations - unconfirmed would be OK.". Agreed, we can do that assuming we still trigger a full wipe off the vision backend for that source. At least until we have a way to retrofit the classifiers objets and val results data structures.

StephenChan:
[...]

PR #290, opened 2020-02-18
Improving label replacement in a source

StephenChan:
To address issue #289.

- On the Add/Remove Labels page, allow removing a label from a labelset when the source only has unconfirmed annotations using that label. Previously, removal was not allowed if there were any annotations, even unconfirmed.
- Change the related management command to not only replace annotations, but also remove the old label from the labelset and initiate a source backend-reset.
- While trying out this functionality, I noticed that the Annotation History page would list a labelset-removed label as "(Deleted label)", even if the label still existed on the site. That seemed misleading, so I changed it.

### DIFF
--- a/project/annotations/management/commands/replace_label_in_source.py
+++ b/project/annotations/management/commands/replace_label_in_source.py
@@ -1,8 +1,10 @@
 from __future__ import unicode_literals
+from datetime import timedelta
 import six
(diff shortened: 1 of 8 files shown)


beijbom:
Thanks @StephenChan . looks good.

PR #291, opened 2020-02-22
Respect the global source-view permission

StephenChan:
There are two types of Permission objects in the database: global permissions and object-level permissions. Global perms include the permission to view sources (i.e. view all sources), or the permission to edit labels (i.e. edit all labels). Object-level perms include the permission to view source 17, or edit source 28.

Up to now, we've only been respecting object-level permissions for source access. That is, in order for user A to view a private source B, one of the following must be true:

- There's an object-level Permission object in the database which specifies 'user A has view access to source B'.
- User A is a superuser (basically bypassing all permission checks).

For @qiminchen's work (and maybe for other situations in the future), we were thinking of adding a third possibility:

- There's a global Permission object in the database which specifies 'user A has view access to any source.' (This could be added, for example, using the admin interface at a URL like `/admin/auth/user/3/change/`.)

------------------

I've concluded that this needs an additional check in Source's `visible_to_user()` in order to work. The good news is that it's basically a one-liner. The not-so-good news is that it's a little unintuitive; if you look at the implementation now, it might seem redundant. Details on what's going on:

- The call `has_perm(perm_name)` checks for a global perm. The call `has_perm(perm_name, obj)` checks for an object-level perm.

- The tricky thing is, `has_perm` behaves very literally: it checks the DB to see if that specific Permission object exists. It can't figure out "hey, this user can view all sources, so they must be able to view source 17 too." This contrasts with django-guardian's `get_objects_for_user()` method, which by default does this more 'semantic' permission checking.

Anyway, besides that, I noticed we were lacking permission tests for some source pages, so I added some. I wanted to write more tests before the Python 3 upgrade anyway.

### DIFF
--- a/project/images/models.py
+++ b/project/images/models.py
@@ -175,7 +175,12 @@ class Source(models.Model):
 
     @staticmethod
     def get_sources_of_user(user):
-        # For superusers, this returns ALL sources.
+        """
+        Get all sources that the user is a member of.
+
+        Special cases due to how permissions work: this returns ALL sources
+        for 1) superusers, and 2) users granted the global source-view perm.
+        """
         if user.is_authenticated:
             return get_objects_for_user(user, Source.PermTypes.VIEW.fullCode)\
                 .order_by('name')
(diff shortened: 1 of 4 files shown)


beijbom:
Looks solid. Thanks @StephenChan . CC @qiminchen .

PR #292, opened 2020-02-23
Fixed spacer export

beijbom:
Updated the spacer export to:

* Ensure the annotations have the same order as the features
* Added info on train and val split
* Added best source robot accuracy and source primary key to meta data.
* PEP8
* Misc scripting cleanup.

### DIFF
--- a/project/vision_backend/management/commands/vb_export_spacer_data.py
+++ b/project/vision_backend/management/commands/vb_export_spacer_data.py
@@ -94,7 +94,7 @@ class Command(BaseCommand):
 
         # Iterate over sources
         for itt, source in enumerate(sources):
-            print("Exporting {}, id:{}. [{}({})] with {} images...".format(
+            print(u"Exporting {}, id:{}. [{}({})] with {} images...".format(
                 source.name, source.pk, itt, len(sources) - 1,
                 source.nbr_confirmed_images))
             if itt < options['skip_to']:
--- a/project/vision_backend/management/commands/vb_export_spacer_data.py
+++ b/project/vision_backend/management/commands/vb_export_spacer_data.py
@@ -18,11 +18,23 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
 
-        parser.add_argument('--min_required_imgs', type=int, nargs='?', default=200,
-                            help="Min number of confirmed images required to include a source.")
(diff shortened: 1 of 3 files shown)


beijbom:
Once this is complete I suggest we do a new, clean export. I can start that as soon as this is merged to master.

beijbom:
@qiminchen : yeah, that is what I do also when I call the feature export script. I can't fully verify until I run this export and test it. Btw. this will all be updated with the new spacer, such that matching is based on actual (row, col) information.

I do think I have to re-export all data. It's only the annotations files that change, but there could be all kinds of changes in the DB, so I rather do a fresh export.

qiminchen:
Sounds great! Then I actually only have to sync the annotations files, info on train and val split and meta data.

beijbom:
@qiminchen : I updated my answer, I think you need to re-sync the whole thing unfortunately.

qiminchen:
Okay, since the order of both features and labels got changed in order to match each other. The good thing is that the data is still correct for the purpose of training base-classifiers.

project/vision_backend/management/commands/vb_export_spacer_data.py
```diff
@@ -43,43 +55,58 @@ def source_meta_json(source):
...
         sources = filter_out_test_sources(sources)
         sources = [s for s in sources if
                    s.nbr_confirmed_images > options['min_required_imgs']]
-        sources = sources[options['skip_to']:]
 
         # Iterate over sources
         for itt, source in enumerate(sources):
-            print("Exporting source id:{}. [{}({})] with {} images...".format(source.pk, itt+1, len(sources),
-                                                                              source.nbr_confirmed_images))
+            print("Exporting {}, id:{}. [{}({})] with {} images...".format(
```
StephenChan: Add a `u` before the opening quote to make it a Unicode string, so that those non-ASCII source names don't trip you up. `print(u"Exporting {}, id:{}. [{}({})] with {} images...".format(`

StephenChan:
Sorry for the delay, forgot to review this. Looks good other than the string issue I just pointed out.

project/vision_backend/management/commands/vb_export_spacer_data.py
beijbom: Thanks! Changed.

PR #295, opened 2020-02-28
added feature inspection crawler

beijbom:
Found a few instances where the extracted features did not align with the content of the DB. In particular the number of features was not the same as the number of points for that image. So wrote this crawler to inspect how common this is.

### DIFF
--- a/project/vision_backend/management/commands/vb_inspect_extracted_features.py
+++ b/project/vision_backend/management/commands/vb_inspect_extracted_features.py
@@ -27,7 +27,7 @@ class Command(BaseCommand):
 
         errors = defaultdict(list)
         for source in Source.objects.filter():
-            print('Inspecting source: {}'.format(source.name))
+            print(u"Inspecting source: {}".format(source.name))
 
             for image in Image.objects.filter(source=source):
                 if not image.features.extracted:
--- a/project/vision_backend/management/commands/vb_inspect_extracted_features.py
+++ b/project/vision_backend/management/commands/vb_inspect_extracted_features.py
@@ -1,0 +1,56 @@
+import json
+import os
+from collections import defaultdict
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from images.models import Source, Image, Point
+from lib.utils import direct_s3_read
+
+from vision_backend.tasks import reset_features
+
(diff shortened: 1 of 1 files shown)


project/vision_backend/management/commands/vb_inspect_extracted_features.py
```diff
@@ -0,0 +1,60 @@
...
+                            type=int,
+                            nargs='?',
+                            default=0,
+                            help="If true, fix erroneous images.")
+
+    def handle(self, *args, **options):
+
+        errors = defaultdict(list)
+        for source in Source.objects.filter():
+            print('Inspecting source: {}'.format(source.name))
```
StephenChan: Unicode again. Either do the same fix as the other PR, or add `from __future__ import unicode_literals` at the top (the latter makes it a bit easier to port to Python 3).

project/vision_backend/management/commands/vb_inspect_extracted_features.py
```diff
@@ -0,0 +1,60 @@
...
+                    assert n_pts == len(feats), \
+                        "n_pts ({}) neq len(feats) ({})".format(n_pts,
+                                                                len(feats))
+                    assert len(feats[0]) == 4096, "Feats dim: ({}) neq 4096".\
+                        format(len(feats[0]))
+
+                    # TODO: once we have the new feature structure from
+                    # TODO: spacer, we can assert that the row, col information
+                    # TODO: matches also.
+                except Exception as err:
```
StephenChan: It's a little safer to change `Exception` to the more specific class `AssertionError`. That way, this except block will only be entered on an assert statement failure, and not some other error with `direct_s3_read` or something. (Assuming that isn't your intention.)

beijbom: Yeah. I wrote it thus since there could be exception due missing file, or incorrect `json` formatting orr other stuff. But in general, I agree that I do tend to over-use this broad catch.

beijbom:
@StephenChan : thanks for the unicode, reminder (again!). Please drop me a line once the code is updated on master.

StephenChan:
@beijbom This is available in production now.

PR #300, opened 2020-03-15
More LIKELY_TEST_SOURCE_NAMES

beijbom:
Mini PR. As I was looking through the spacer exports, I found these sources:
```
[15/03/2020 12:51:54]: Exporting Reviewer 1 Data, id:224. [40(333)] with 579 images...
[15/03/2020 12:56:36]: Exporting Reviewer 2 Data, id:225. [41(333)] with 579 images...
[15/03/2020 13:01:16]: Exporting Reviewer 3 Data, id:226. [42(333)] with 579 images...
[15/03/2020 13:05:56]: Exporting Reviewer 4 Data, id:227. [43(333)] with 579 images...
[15/03/2020 13:10:40]: Exporting Reviewer 5 Data, id:228. [44(333)] with 578 images...
```
They weren't caught by the current `LIKELY_TEST_SOURCE_NAMES` options, so I added one more string.

### DIFF
--- a/project/config/settings/base.py
+++ b/project/config/settings/base.py
@@ -629,7 +629,8 @@ SELENIUM_TIMEOUTS = {
 }
 
 # We filter on sources that contains these strings for map and some exports.
-LIKELY_TEST_SOURCE_NAMES = ['test', 'sandbox', 'dummy', 'tmp', 'temp', 'check']
+LIKELY_TEST_SOURCE_NAMES = ['test', 'sandbox', 'dummy', 'tmp', 'temp', 'check',
+                            'reviewer']
 
 # NewsItem categories used in NewsItem app.
 NEWS_ITEM_CATEGORIES = ['ml', 'source', 'image', 'annotation', 'account']


beijbom:
@qiminchen : FYI: I deleted these from the beta_export bucket, let's not use them for your work.

qiminchen:
okay

StephenChan:
Hmm, did the sources themselves get deleted as well? I don't see them. Actually, it seems are no more sources containing 'reviewer', in which case this PR wouldn't be needed.

beijbom:
Yeah. I deleted them. They were some very old sources I created for some
academic reviewers. Ok, sure, we can close this then.

StephenChan:
Ah, okay. If you think you might create similar sources in the future again, we can merge it, else close it.

kriegman:
Oscar,

Looking at this, I think it's really important.

Up to now, Qimin has been training the network using sources 225, 226, 227,
228

and had included source 224 in the test set.

My guess from your concern is that these 5 sources are all the same image
data, perhaps labelled almost the same , maybe by different humans or
machines.

Qimin was getting extraordinary accuracy on S224 (95%-99%).
This would explain why he was getting this accuracy -- testing on the
training set.

So, he should computely eliminate 225, 226, 227, 228 from training and 224
from testing.

Make sense?

Fortunately, s224 is the only one with extraordinarily high accuracy. So,
it is probably the only one which has this problem.

David

beijbom:
@kriegman : Yes, these sources were all **identical**. I created them for reviewers of one the CoralNet papers. They contain subset of the Moorea LTER data. So, yes, they should be excluded. I think it's fine to keep them along for this round of hyper parameter tuning, but for the final training, let's get rid of them. Glad that explains the outlier.

Issue #284, opened 2020-02-03
Spacer SWE and dev-ops.


beijbom:
This ticket is to do SWE and dev-ops for spacer. This includes:

- [x] Create a Dockerfile for better container building process. Support both caffe and pytorch inside the container. 
- [x] Test new container to ensure backwards compatibility.
- [x] Setup Travis CI on GitHub.
- [x] Isolate Caffe such that the rest of repo functions even when Caffe is not available. 
- [x] Make and publish PIP package. This would only suppor the PyTorch. Make sure the PIP package supports both python 2.7 and 3.7.
- [x] Better Modularization of codebase. Abstract away the specifics of the classifiers and feature extractors to allow implementation of new versions.
- [x] Document messages and "API".
- [x] Abstract away storage and support both local and S3.

beijbom:
@StephenChan @qiminchen : FYI: the more I looked at the spacer codebase, the more I realized I need to do some dev-ops and SWE before we add the new feature extractors. This ticket lists the main todos related to this. My goal is that we continue supporting the current feature extractor.

StephenChan:
Okay. If you feel that you're short on time to get all these tasks done, let us know if there's something we can help with.
By SWE you mean software engineering, right?

beijbom:
Hey @StephenChan .  Yeah, Software engineering. I think I got this one. I have done enough thinking on this that it should be relatively smooth sailing. 

The one thing that would help is if I don't have to provide support for both python 2.7 and 3.{6, 7} in the pip package. I haven't looked at it yet, so maybe it is fine, but I do anticipate a bit of headache to get that going. So if you feel ready to move CoralNet to python 3.6 or 3.7, and that you can do it soon, then that would help. If not - and it turns out to be tricky to support both - I'll start with 2.7 only.

StephenChan:
> The one thing that would help is if I don't have to provide support for both python 2.7 and 3.{6, 7} in the pip package. I haven't looked at it yet, so maybe it is fine, but I do anticipate a bit of headache to get that going. So if you feel ready to move CoralNet to python 3.6 or 3.7, and that you can do it soon, then that would help. If not - and it turns out to be tricky to support both - I'll start with 2.7 only.

[...]

Issue #157, opened 2018-06-03
CPC export: more unit tests


StephenChan:
- Import then export, should be the same.
- Import, then change something, then export.
- All images in search have prev-uploaded CPCs.
- Some images in search have prev-uploaded CPCs.
- No images in search have prev-uploaded CPCs.
- All possible annotation area types.
- CPC prefs form's initial values.
  - No exports/uploads at all yet, should be blank.
  - Form is blank, then you fill it in when doing a CPC export, then you go to do another CPC export and these values should be as you entered before.
  - Done an upload before, values should be filled in.
- Selenium or QUnit: the form should be hidden or not based on prev-CPC status.

StephenChan:
Some of this was done in de5b39440e4d94a7445fd75381936b3a4ae42378 (September 2018).

> Import, then change something, then export.

Covered by `test_export_with_previous_cpc_for_this_image`

> All images in search have prev-uploaded CPCs.
> Some images in search have prev-uploaded CPCs.
> No images in search have prev-uploaded CPCs.

Covered by:

- `test_form_init_when_all_images_in_search_have_cpcs`
- `test_form_init_when_some_images_in_search_have_cpcs`
- `test_form_init_when_no_images_in_search_have_cpcs`
- `test_form_init_when_no_cpcs_ever_uploaded`
- `test_export_with_no_cpcs_saved`
- `test_export_with_previous_cpc_for_this_image`

> CPC prefs form's initial values.

Generally covered by the `CodeFileAndImageDirFieldsTest` class.

> Import then export, should be the same.
> All possible annotation area types.
> Selenium or QUnit: the form should be hidden or not based on prev-CPC status.

Not covered yet. I'll do the first two soon. The third one probably would be best as a Selenium test though, and I'd prefer not to deal with those at the moment. Might move that to another issue.

StephenChan:
> Import then export, should be the same.
> All possible annotation area types.
> Selenium or QUnit: the form should be hidden or not based on prev-CPC status.

Did the first two, and moved the third one to #305. (Although it turns out the third one kind of has a test already; it's just a bit unwieldy, checking for exact CSS.)

Issue #155, opened 2018-06-03
Unit test / edge case checklist (2020/04)


StephenChan:
Miscellaneous unit tests and edge cases which aren't covered by other issues. Might split some of these off into their own issues later.

- accounts
  - [x] login: Test redirect after login
  - [x] profile list: Test pagination
- annotations
  - annotation tool: Load the annotation form on the annotation tool page...
    - [x] with blank annotations
    - [x] with all unconfirmed annotations
    - [x] with all confirmed annotations (should show 'All Done')
    - [x] with some annotations confirmed, some blank
    - [x] with some annotations confirmed, some unconfirmed
    - [x] each of the above with 'show machine annotations' setting on or off
  - [x] annotation tool: Load the settings form and ensure the user's saved values are present
  - [x] save annotations ajax: Save annotations after changing some from unconfirmed to confirmed
  - [x] is annotation all done ajax: If some are unconfirmed, should not be all done
  - annotation area edit
    - [x] Load a percentage based annotation area
    - [x] Load a pixel based annotation area
    - [x] Submit an edited annotation area
  - point generation
    - [x] Stratified random and uniform grid, not just simple random.
- Browse Images
  - [x] Handling of invalid search parameters
  - [x] Correct indication of image statuses (border color, or presence of CSS class)
  - Asychronous generation of thumbnails as needed (probably needs some Selenium tests)
    - [x] Load thumbnails which already exist
    - [x] Generate thumbnails which don't exist yet
- [x] Image real-filename generation (for saving to S3): ensure no two images can have the same filename. Test by replacing the filename generation function (using Mocker?) with something that only can generate a few unique filenames.
- [x] Caching of label popularity. Fetch the popularity once, then replace the full popularity implementation with something that raises an error, then fetch popularity again.

StephenChan:
Going to tackle some of these items soon and leave others for later.

StephenChan:
Okay, time to clean this up. Here are the items I'm discarding or putting off for now:

>       * annotation area edit
>         * QUnit tests
>           * [ ]  Change each of the 4 text fields
>           * [ ]  Drag the resize handles; ensure the drag extremities are where they should be
>           * [ ]  Text field errors: non-integers, beyond min/max, min greater than max

Moved to issue #305.

>       * point generation
>         * [ ]  Unusual annotation areas: min and max very close or the same, and decimal percentages.
[...]

Issue #77, opened 2016-10-12
Write tests for image details page


beijbom:
There are several operations possible on the image detail page. For example: re-generate points, set annoatation-area to source default; remove all annotations etc. None of these have any tests.

StephenChan:
e08afaa0164425fc16ae4ed60841d70f2eff59a6 adds tests for these operations.

PR #301, opened 2020-03-19
Pre-Python 3 unit test roundup

StephenChan:
Additional unit tests before working on Python 3 compatibility.

This particular PR will make very minimal changes to non-test code, so that it's relatively safe to merge. Once this is done, I will start the Python 3 feature branch as discussed in issue #284.

For this PR, I'm thinking of aiming for:

- Issue #54; should already be covered by a commit here
- Issue #77
- Issue #157, since it's close to done
- Parts of issue #155, more focusing on common functionality than edge cases
- Other common functionality that's lacking tests according to `coverage`


beijbom:
@StephenChan : really nice to see this WIP PR. I think it sounds like a great effort.

StephenChan:
All right, I think I'm ready to move on from testing, unless I messed something up.

- #54, #155, and #157 have been dealt with. (Also #77 in the earlier security PR.) I moved the skipped items to other lower-priority issues.
- Most of the TODO comments in the test .py files have been addressed. Again, I skipped some.
- Two non-test behaviors were changed:
  - Filename generation for uploaded images has been changed a bit. It turns out that this process was already guarded from generating duplicates, but it would accept instances of one image being named `1234567890.png` and another image being named `1234567890.jpg`, which could be confusing, even if not actually dangerous. I changed the code to minimize the possibility of this happening, and if it does ever happen, we'll get an email. See `images/models.py`.
  - The `label_new_ajax` view now responds with JSON instead of HTML when the user isn't logged in. See `labels/views.py`. This is a corner case since you have to be on the labelset add/remove page to submit to this view normally, and that page requires login.
- Code coverage is up from 91% to 93%. Not that big of a jump, but it's something. Most of the remaining un-covered lines fall into a few tricky categories:
  - Backend tasks and functions related to backend_overview
  - Statistics page (it's still broken, of course)
  - Selenium and other tests that are skipped for now (yes, tests themselves are counted in the coverage)
  - Migration if/else cases which check for existing data (we'll squash most of these files sooner or later anyway)

StephenChan:
...As I'm starting to look at Python 3 porting now, I should note: I may still add more tests for things that come to mind during Python 3 porting. But this particular PR has enough stuff in it.

Issue #307, opened 2020-04-05
Consider overriding SITE_DIR setting for unit tests


StephenChan:
The biggest reason is that log files are currently shared between the actual server and unit tests. We typically only run unit tests in dev environments (and maybe staging), but it'd still be nice to not clog up the log files with unit-test output. If unit tests use a separate `SITE_DIR`, then that directory could be cleaned up after testing.

Other files located based on `SITE_DIR` include `tmp` files, and the maintenance-status file (from issue #155). We might also be able to simplify the `CACHES` setting override, which looks like this right now:

```python
test_settings['CACHES'] = {
    'default': {
        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
        'LOCATION': 'test',
    }
}
```

StephenChan:
In PR #324, we had to mute the logger during testing, perhaps because of how spacer uses the logging module. So if we still want the logger activated during unit tests (for example, to test the log output itself), then that is something that needs resolving.

StephenChan:
Just noticed that with the recently added views/tasks logs (e.g. `coralnet_views_tasks_debug.ssv`), unit tests can fail if a development server is running at the same time, getting an error like `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process`. This error shouldn't occur in this situation if this issue's idea is implemented.

Issue #59, opened 2016-09-25
Upgrade codebase from Python 2.7.x to Python 3.x


StephenChan:
## Motivation: End of support for Python 2, in both Django and Python itself

* Python 2.x end of support is [January 1, 2020](https://pythonclock.org/). It won't get security updates after that.
* Django 1.11.x is the last version to support Python 2.x. Django 1.11.x's end of support may be as early as [April 2020](https://www.djangoproject.com/download/). Django 2.0 onward only supports Python 3.x.
* In addition to Django dropping Python 2 support, we can expect other third-party Python packages that we use to drop it as well.

## Which version to target

Python 3.6.

Python 3.4's end of life is [March 2019](https://docs.djangoproject.com/en/dev/releases/2.0/), even earlier than Python 2.7's end of life. 3.6.x is the latest stable version as of 2018.05.11. See other versions' end-of-life dates [here](https://devguide.python.org/#status-of-python-branches).

So from there the question is 3.5 or 3.6 based on third-party package support. As it turns out (#185), all of our third-party packages support 3.6, so we should aim for that.

## How to update our codebase

The trickiest part is Unicode strings vs. byte strings. Fixing our code for that often requires a good semantic understanding of the code we're looking at, so it's tough to just go through the whole codebase at once and fix that. I propose that we gradually make this change while working on other features like the API.

For other 2-to-3 changes, it's probably easier to change the whole codebase at once so we don't have to keep those changes in mind anymore. But... we should write more unit tests first.

- Details on 2 vs. 3 differences
  - https://docs.python.org/3/howto/pyporting.html
  - http://python-future.org/compatible_idioms.html
  - Use the `six` module to write code compatible in both Python 2 and 3. See the [six docs](https://pythonhosted.org/six/).
    - I'm not actually using the `future` package since it doesn't offer that much more benefit over six, IMO, and PyCharm inspections didn't seem to play nice with future when I tried it.
- Also see the changelogs for later 3.x versions, but don't use 3.x-only features yet (that's for #186)
- Use the PyCharm inspections which check Python version compatibility.
- Ctrl+F our project for mentions of "Python 2", "Python 3", etc. in comments. I left at least a couple of comments like that.
- Starting in Django 1.10, Unicode usernames are supported when using Python 3.x. We must decide whether to allow Unicode usernames or add our own validator to ensure it's ASCII only. (Most likely we'll stick with ASCII usernames for simplicity.)
[...]

Issue #302, opened 2020-03-23
Setup up new scalable cluster on AWS with new spacer.


beijbom:
To make transition to new spacer backend, and python 3 shakeout, setup a new ECS and new SQS queues, etc.  This ECS will then be used in production and we can retire the old one.

beijbom:
This is being worked on in https://github.com/beijbom/pyspacer/pull/17. Proof of concept is up and running.

beijbom:
This is done now.

Issue #264, opened 2019-12-02
vision_backend app cleanup & testability


beijbom:
This issues is to do some SWE on the vision backend app. In particular
- Remove any unnecessary dependency.
- Increase testability by implementing a testable backend.
- Make sure all methods are tested as suggested in 
https://github.com/beijbom/coralnet/issues/182
- Change to new spacer.

beijbom:
Will continue working on this after https://github.com/beijbom/coralnet/issues/284 is complete.

beijbom:
The rest of this ticket is tracked in https://github.com/beijbom/coralnet/pull/323.

beijbom:
Closing this after the added tests.

PR #309, opened 2020-04-12
Python 3 compatibility, 'safer' changes only

StephenChan:
Made a lot of progress this week towards #59. This PR should do the following:

- Get the server running and all tests passing in both Python 2.7 and 3.6.
- Achieve full Python 3.6 compatibility in test-covered files.
- Make only basic compatibility changes in non-test-covered files, including:
  - Most of `vision_backend/tasks.py` and `vision_backend/task_helpers.py`. I suppose these may be largely rewritten for the pyspacer rollout anyway (@beijbom: can you confirm?).
  - Most management commands, since they either use the VB tasks or are tricky to test for some other reason (e.g. they use `SITE_DIR` directly; see #307).
  - All miscellaneous scripts: `images/scripts.py`, `vision_backend/scripts.py`, and contents of the `scripts` folder at the repo root. I didn't write tests for these because I don't know which ones have been used anytime recently.

'full Python 3.6 compatibility' just means it works in Python 3.6 (and still 2.7) as far as I can tell, and the code isn't particularly ugly for Python 3.6 purposes. I could have added a few more imports like `from six.moves import range`, `from __future__ import unicode_literals`, etc. to really make sure 2.7 ran exactly the same way as 3.6, but to me it seemed unnecessary if the tests already passed on both. Just as long as we don't write new untested code from now to the pyspacer rollout, and I don't think we plan to.

'only basic compatibility changes' means not touching the Unicode strings vs. byte strings stuff, for the most part. Just things like fixing print statements, fixing some easy imports, maybe division ops if they look straightforward enough.

Highlights in the `vision_backend` code:

- I added more tests for `confmatrix.py` and `views.py`, and also manually tested `backend_overview` in Py2 and Py3.

- It seems that numpy's integer types are descendants of the standard `int` type in certain numpy builds, but not others. If they don't descend from `int`, that means numpy's integer types can't be used directly for indexing into querysets. Hence the extra `int()` casts here and there.

Third-party package notes:

- I removed the scipy and scikit-learn dependencies, since they weren't being used as far as I could tell. Removing the deps from coralnet means that our virtualenv can simply match the scipy/scikit-learn requirements in pyspacer.

- I haven't updated the numpy version requirement yet, but I've actually been running a later version of numpy for Py3.6 testing on my machine, because numpy 1.11.1 didn't have a Py3.6 binary for me to install. In fact, the tests pass for me on numpy 1.18.1 - same as the version in https://github.com/beijbom/pyspacer/pull/10. If I'm not mistaken, all of our numpy usages are actually test-covered now, though I still wasn't sure if bumping the version could be considered a 'safe' Py3 compat move. The aforementioned integer-type problem was kind of subtle, after all.

- All the other third-party package versions are untouched. They just seemed to work in Py3.6, eerily enough!

By my judgment at least, if this PR passes review, it should be fine to merge now. Again, still Py2.7 compatible.

If I didn't forget anything, we mainly have these 'higher-risk' Python 3 compatibility changes left:

- Full Py3 compatibility for `vision_backend` (maybe wait until tasks are reworked?)
- Full Py3 compatibility for commands/scripts (@beijbom: which ones are still being used recently? I can prioritize those.)
- Updating numpy (Edit: there's another wrinkle to this - 1.17+ is Py3 only)

I can work on these items in another branch. Then we'd have the choice of merging those changes along with the rest of the pyspacer rollout, or merging them earlier. I'm fine either way.


beijbom:
@StephenChan : sorry, I must have missed this. Is this PR ready for review?

StephenChan:
No worries. Yeah, it's ready for review.

beijbom:
@StephenChan : I took a look at this. Looks good in general. I added `python 3.6` to Travis and all tests pass there as well as on my local dev. system.  I also bumped `tqdm` package a bit to be compatible with `pyspacer`. 

- Re. management commands: I actually use all of them still. So if you can check those, that'd be great.
- Re. the vision backend module. It is up to you. My thinking is that we will add the spacer dependency, and rework that module at the same time. Note, however, that this wont be python2 compatible, so we will have to either bump the production server to Python 3.6 first or do this on a feature branch. I suppose the same goes for the numpy update. 

- As a side-note: I tried installing `pyspacer` and importing it. It is taking a long time to import. It seems it's related to how I check connections in `config.py`. I'll take a look at that over in the spacer repo.

beijbom:
@StephenChan : please ignore my last comment regarding spacer. It was related to my SSL certificates on python3.6 on OSX. I found a way to resolve it. https://stackoverflow.com/questions/42098126/mac-osx-python-ssl-sslerror-ssl-certificate-verify-failed-certificate-verify

StephenChan:
> I added `python 3.6` to Travis and all tests pass there as well as on my local dev. system. I also bumped `tqdm` package a bit to be compatible with `pyspacer`.

Sounds good.

> * Re. management commands: I actually use all of them still. So if you can check those, that'd be great.

Okay. How about the miscellaneous scripts: `images/scripts.py`, `vision_backend/scripts.py`, and contents of the `scripts` folder at the repo root?

> * Re. the vision backend module. It is up to you. My thinking is that we will add the spacer dependency, and rework that module at the same time. Note, however, that this wont be python2 compatible, so we will have to either bump the production server to Python 3.6 first or do this on a feature branch. I suppose the same goes for the numpy update.

All right. So I guess we pick what we want to deal with:

- If we bump production to Py3.6 first, the Beta-1 `vision_backend` will need one last round of debugging for those Py3.6 compat changes, even though we plan to rework `vision_backend` very soon per issue #264.

- If we keep the rest of the Py3.6 compat changes in a feature branch, and roll that out at the same time as #264, we'll be trying Py3.6 in production for the first time while also getting it on pyspacer for the first time.

At this point I'm leaning toward the feature branch idea that you suggested. Hopefully I'm not getting complacent, but I'm feeling reasonably good about the state of the non-VB code and tests. If I can get this PR onto production now (still running Py2.7) and take care of any problems that crop up, I think the actual production switch to Py3.6 shouldn't be too bad. Compared to that, I wouldn't be particularly looking forward to debugging the Beta-1 VB, to be honest.

beijbom:
@StephenChan : Ok. let's do the feature branch then. And to be clear: we can make it very small: basically just https://github.com/beijbom/coralnet/issues/264 and perhaps some other minor cleanups related to dropping support for 2.7. 

I'll start that once https://github.com/beijbom/coralnet/issues/302 is done. (no strict dependency; I just want to do that first).

If you fix the management commands in this PR, you can leave the rest of the misc scripts for me to clean up later.

StephenChan:
> Ok. let's do the feature branch then. And to be clear: we can make it very small: basically just #264 and perhaps some other minor cleanups related to dropping support for 2.7.
> 
> I'll start that once #302 is done. (no strict dependency; I just want to do that first).

Okay, cool.

> If you fix the management commands in this PR, you can leave the rest of the misc scripts for me to clean up later.

For the remaining management commands, I was actually thinking of fixing them in the feature branch instead of this PR. Since most of the remaining commands use the VB tasks.

Sounds good on the misc scripts, though. (I know I wrote s3_sync, so you can leave that one for me.)

beijbom:
> For the remaining management commands, I was actually thinking of fixing them in the feature branch instead of this PR. Since most of the remaining commands use the VB tasks.

sounds good.

StephenChan:
Merged and now in production.

Issue #314, opened 2020-05-02
Resolve CoralNet and Spacer 0 vs 1 indexing misalign.


beijbom:
Unfortunately, we have a situation where CoralNet uses 1-indexed `row, col` locations while spacer uses 0-indexed. This thread is to discuss ways to mitigate. 

First: why is this a problem? Well, the new spacer is much better at tracking `row, col` locations throughout the process, and I was meaning to double check against that at every step when parsing return messages and data structures. We also store `row, col` in the actual `.json` feature files which are part of the CoralNet s3 blob store. 

So what can we do? Well, there are three options.

**1) write converters:** The most obvious option is to write converters at every interaction point between spacer and Coralnet. These include at least:
- Submitting features for extraction.
- Reading feature extraction return.
- Submitting a new training job.
- Submitting and parsing classification jobs.

Also: the [stored features](https://github.com/beijbom/pyspacer/blob/master/spacer/data_classes.py#L142) for an image would use zero-indexing. So it's not just the messaging code that needs to be aware of this, we would be creating a permanent records using two different conventions.

**2) Change spacer to use 1-indexing:** This is probably the easiest to achieve, by, for example subtracting 1 from center [here](https://github.com/beijbom/pyspacer/blob/5158bf1e6f116fe70ff06ed71705326f73c44dfe/spacer/extract_features_utils.py#L54), and also update the [check_rowcols](https://github.com/beijbom/pyspacer/blob/5158bf1e6f116fe70ff06ed71705326f73c44dfe/spacer/task_utils.py#L6) method.

What I don't like about this is that 0-indexing is the "correct" indexing, in the sense that is it used by most (all) python libraries such as PIL,  numpy, in addition to python itself.

**3) Change CoralNet to 0-indexing:** This is probably more work and would include
- database migration. This one seems straight-forwards since it just one table.
- Update code that generates row/cols to use 0-indexing. E.g. [here](https://github.com/beijbom/coralnet/blob/master/project/annotations/model_utils.py#L96
).
- Update consumer code, such as renderings and cropping to use 0-indexing.
- Update import from CPCee and in general make it clear for consumers that we use 0-indexing. (Well, we could of-course use 0-indexing internally but 1-indexing when exporting, but that might be another invitation for trouble)

On the "Update consumer code" topic, I looked around a bit, and it seems some of the  consumer code is already 0-indexed, notably 
[...]

PR #318, opened 2020-05-14
Consistently start Point rows/columns at 0, not 1

StephenChan:
For issue #314.

- Added a data migration to convert existing points and annotation areas to 0-indexing.

    - The annotation areas defined by pixel counts, not percentages, need to be converted.
    
    - As of a few days ago, there were no out-of-range point locations in production (assuming 1-indexing). I'll double-check again before running this migration in production though.
    
    - The migration runs in 4 minutes in staging. Staging has 10m annotations, and production has 50m annotations, so I'd budget maybe 30 minutes for running the migration in production.

- Updated the following to use 0-indexing:

    - Point generation.
    - Individual-image annotation area editing.
    - `removepointoutliers` management command. Also sped up the command with F expressions.
    - Bounds checks in annotation upload by CSV.
    - Annotation upload by CPC.
    - Annotation export to CPC.
    
- Updated patch generation and annotation tool code as discussed in issue #314. To recap, patch generation was off by half a pixel, while annotation tool code had a mix of indexing schemes and another half-pixel issue.

- The API was already on 0-indexing, even the docs. So no changes there. (Yes, this means I was confused on the indexing as recently as a couple months ago.)

- Added row/column bounds checks to the `Point` model's `save()` method.

Let me know if you can think of any other row/column code to double check.


project/images/models.py
```diff
@@ -780,6 +790,20 @@ def machine_confidence(self):
             # No scores means max(scores) will raise this
             return 0
 
+    def save(self, *args, **kwargs):
```
beijbom, line 793: nice to see this here.

project/visualization/utils.py
```diff
@@ -111,33 +111,42 @@ def generate_patch_if_doesnt_exist(point_id):
...
+    approx_region_size = int(max(image.original_width, image.original_height)
+                             * settings.LABELPATCH_SIZE_FRACTION)
 
-    # Load the image convert to RGB.
+    # Load the image and convert to RGB.
     im = PILImage.open(original_image_file)
     im = im.convert('RGB')
 
     # Crop.
+    # - Both CoralNet coordinates and Pillow coordinates start from 0 at the
```
beijbom, line 130: this part looks good also.

beijbom:
I only reviewed some of it, but the stuff I was thinking about are implemented. That check on the model makes me happy.

StephenChan:
Thanks for looking at this @beijbom . I'll merge into master and rebase the `beta2-rollout` branch. I haven't written the blog post for these changes yet, but I'll certainly do that before updating production.

Issue #200, opened 2018-12-19
JScolor plugin's images get 404 errors


StephenChan:
I was looking at the nginx logs and saw something like this:

`"GET /image/871828/annotation/tool/jscolor/hs.png HTTP/1.1" 404`

This image URL is supposed to be something like `/static/jscolor/hs.png` instead. Maybe it's being mistakenly defined as a relative URL like `jscolor/hs.png`, which is then appended to the end of the current page's URL.

Indeed, I saw that at least one of the color picker's images doesn't show up on the production site.

StephenChan:
Done in PR #321.

Issue #319, opened 2020-05-21
Annotation tool: zooming and rotation support are buggy in many browsers


StephenChan:
The annotation tool uses CSS absolute positioning on the image. This positioning scheme is buggy on Chromium-based browsers (Chrome, Edge, Brave, and others) and Safari. The behavior is buggy in two ways:

1. Image position is wrong after zooming in a few times: https://groups.google.com/forum/#!topic/coralnet-users/QZvBTcQj8Oo

2. Image position is wrong if the image has rotation metadata (or maybe just a vertical aspect ratio?), no zooming required: https://groups.google.com/forum/#!topic/coralnet-users/A34j7B-4OqE

Chromium absolute-positioning bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1001438

- This bug report came up around the same time as the reports of the annotation tool bug. Judging from the comments, it's not 100% clear if the issue was actually fixed. I'm also not sure if this is actually the same issue as what the annotation tool is seeing. In any case, it's been several months since the bug report was closed, and we are still seeing the annotation tool problem. So, regardless of whether this is a browser bug or our bug, it seems we need to make our own workaround.

Report that Safari has the bug too, at least for the image rotation: https://groups.google.com/forum/#!topic/coralnet-users/ySgW_P0gY6s

Currently, the only user-side workaround is to use Firefox. But that's not the main browser for a lot of people these days, so we should try to be better than that.

Two actions to take here:

- [x] Short-term (Beta 2): Update the annotation tool help dialog's browser compatibility table with info on this issue. It still says Chrome and Safari are 'OK'. (Done: PR #321)

- [ ] Long-term (Beta 3): Rework the image positioning code so that it works cross-browser. This could be a significant refactor with issue #55 (anno-tool unit testing) as a prerequisite.

StephenChan:
If I'm not going crazy here, the rotation issue might also be present in Firefox now.

StephenChan:
The zoom bug now has a fix in PR #329.

Regarding rotation: for some reason, I can't currently replicate any browser differences here, even without the above fix. I wonder if it's because of the very recent Firefox/Chrome changes I just wrote about in issue #107.

StephenChan:
I'll give a wrap-up on the Chromium rotation bug in #107, since it was actually the same as one of the rotation bugs there.

Here's a wrap-up on the zoom problem:

## Zoom problem in Chromium browsers

[...]

Issue #184, opened 2018-06-12
Use npm to track Javascript plugin versions


StephenChan:
It's a pain to track down these plugins individually. Having a unified way to track them, similar to pip for Python packages, could save a fair bit of effort in the long run.

Do this after #183, since some of our current package versions were picked up before they started supporting npm. This means switching to npm first would update multiple packages at once, making it harder to debug things.

npm support among the packages we use:

- jQuery: supported
- jQuery UI: supported
- bootstrap: supported
- highcharts: supported
- mousetrap: supported
- markerclusterer: old version only, as `marker-clusterer-plus` on the npmjs website. [This repo](https://github.com/googlemaps/v3-utility-library) is the up to date one. But [this issue](https://github.com/googlemaps/v3-utility-library/issues/427) shows that this repo is not on npm yet. I think we can just use the old version so that it plays nice with npm.
- jscolor: GitHub only, at [this repo](https://github.com/EastDesire/jscolor), not on the npmjs website. [Here's how to use with npm](https://stackoverflow.com/questions/17509669/how-to-install-an-npm-package-from-github-directly).

One solution for having npm and Django work together: the [django-npm](https://github.com/kevin1024/django-npm) package. Using this, a run of `collectstatic` would collect from the `node_modules` folder to `static/[NPM_STATIC_FILES_PREFIX]/[filename.js]`.

We need to figure out the best way to install and set up npm in production and in development, and add that to the docs.

PR #321, opened 2020-05-22
Minor annotation tool fixes

StephenChan:
- Update the help dialog's compatibility info, particularly to mention the point positioning issue with Chromium-based browsers and Safari (detailed in issue #319).
- Make the size setting fields bigger on Firefox. The numbers were getting cut off. As the comment explains, the code can be better, but it'll work for now.
- Update the jscolor plugin so that it no longer relies on image resources (which were failing to load with DEBUG = False); fixes issue #200.

The old jscolor version loaded image resources by assuming they were in the same directory as `jscolor.js`; specifically, it would search the page for a `jscolor.js` script element, and parse its `src` attribute with regex to get the directory path. But since we have something in Django that turns it into a `jscolor.<hex hash>.js` script element, jscolor wasn't able to find the expected element. The new jscolor version does not come with any image files, and just draws the color picker gradient on the fly with an HTML canvas.

I also ended up applying a hack (custom code change without a proper package-fork) to the new jscolor, as detailed in the 5th commit. I went ahead with that partially because we may want to replace jscolor in the near future anyway (as I'll talk about shortly in issue #184).


kriegman:
I've got a bunch of things scheduled, but could meet at 12:30?
Does that work?

David

StephenChan:
@kriegman We didn't call for a meeting.

Issue #282, opened 2020-01-21
IOError at /label/<id>/example_patches_ajax/ - image file is truncated (<number> bytes not processed)


StephenChan:
I guess we don't have an issue for this error yet, but I'm pretty sure it's been happening for a while. And fairly frequently, too.

Example after issue #275 was resolved: https://coralnet.ucsd.edu/admin/errorlogs/errorlog/2438/change/
Image in question: image ID 68379
Happened at the `im = im.convert('RGB')` line.

Example before issue #275 was resolved: https://coralnet.ucsd.edu/admin/errorlogs/errorlog/2069/change/
Image in question: image ID 1192943
Happened at the `region = image.crop(box)` line.

This link describes a similar error message: https://stackoverflow.com/questions/12984426/python-pil-ioerror-image-file-truncated-with-big-images

beijbom:
I think the problem is that `im.convert` requires image metadata for PIL to know what color space the image is in. What I suggest is that we catch the `IOError` and then run a "poor mans" conversion, similar to what we do in spacer. That may not handle the most exotic color spaces correctly, but it will do something reasonable in most cases. 


```
_ = np.array(im)  # For some images np.array returns an empty array.
im = np.array(im)  # Running it twice fixes this. Don't ask me why.

if len(im.shape) == 2 or im.shape[2] == 1:
   im = gray2rgb(im)
im = im[:, :, :3]  # only keep the first three color channels

def gray2rgb(im: np.ndarray) -> np.ndarray:
    """
    Convert gray image to RGB image
    :param im: gray image to be converted
    :return: RGB image
    """
    w, h = im.shape
    ret = np.empty((w, h, 3), dtype=np.uint8)
    ret[:, :, 0] = im
    ret[:, :, 1] = im
    ret[:, :, 2] = im

    return ret
```

PR #322, opened 2020-05-25
fixed missing metadata issues

beijbom:
Fixes https://github.com/beijbom/coralnet/issues/282


project/visualization/utils.py
```diff
@@ -1,16 +1,21 @@
...
+from PIL import ImageFile
 from django.conf import settings
 from django.core.files.storage import get_storage_class
 from django.db.models import Q
-import django.db.models.fields as model_fields
-from PIL import Image as PILImage
+
 from images.models import Point, Image, Metadata
 
+ImageFile.LOAD_TRUNCATED_IMAGES = True
```
beijbom, line 17: @StephenChan : I added this here, but perhaps better to move to project level config? (`config/settings/base.py`)?

beijbom:
@StephenChan : I got tired of seeing these error messages to figured I'd fix it. Was easier than I expected although can't be sure works for all images.

project/visualization/tests/test_utils.py
```diff
@@ -1,16 +1,18 @@
 from __future__ import unicode_literals
+
+import os
 from io import BytesIO
+
 import mock
 from PIL import Image as PILImage
 from PIL.Image import SAVE as PIL_SAVE
-
+from django.conf import settings
```
beijbom, line 9: These import changes are due to using the pycharm auto-format imports.

StephenChan: Works for me.

project/visualization/utils.py
StephenChan: Moving to `config/settings/base.py` seems better, yeah. I just tried that and it works, so I'll add that change to this PR.

StephenChan:
Thanks @beijbom , that's a nice find. I'll just push a commit to move the LOAD_TRUNCATED_IMAGES line to settings, as mentioned in my inline comment.

PR #327, opened 2020-06-05
Updated CoralNet support

kriegman:
Added Pacific Blue Foundation


beijbom:
Looks good to me. Does this include the grant with Andreas?

PR #329, opened 2020-06-09
Annotation tool - Zoom and rotation fixes

StephenChan:
Zoom fix was much easier than expected. That's one part of issue #319.

But for the more serious rotation problem, I can't even replicate the browser differences anymore. So, not sure what other quick fix(es) to include in this PR yet, if any.

### DIFF
--- a/project/annotations/static/js/AnnotationToolHelper.js
+++ b/project/annotations/static/js/AnnotationToolHelper.js
@@ -184,7 +184,13 @@ var AnnotationToolHelper = (function() {
         imageTopOffset = parseFloat(imageTopOffset.toFixed(3));
 
         // Set styling properties for the image canvas.
+        //
+        // We need to set both the width and height. If we set only one
+        // dimension, the browser MAY auto-scale the other dimension, but it's
+        // not guaranteed (e.g. with very large sizes in Chromium-based
+        // browsers).
         $(getImageCanvas()).css({
+            "width": imageDisplayWidth,
             "height": imageDisplayHeight,
             "left": imageLeftOffset,
             "top": imageTopOffset,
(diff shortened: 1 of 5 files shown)


beijbom:
Wow. That was easy indeed, although only easy if one knows where to look.

kriegman:
Nice.

StephenChan:
Added the fix for the recent browser-independent orientation problem: edit the annotation tool's in-memory image to remove its EXIF orientation before display. I tried some odd JPEG metadata cases and they worked fine. If there's an unexpected problem with this code, a Javascript alert should pop up letting the user know that something is wrong.

The plugin I chose, piexifjs, doesn't currently recognize EXIF data in PNG files, and so I wasn't able to edit the image EXIF for the PNG case. But none of the browsers today seem to recognize EXIF orientation in PNG anyway (indeed, EXIF was only added to the PNG spec in 2017). So this will do for the time being, but it's worth noting that PNG EXIF will probably have to be supported later.

Next steps for me here:

- Confirm whether the Chromium-specific orientation problem is fixed or not.
- Update the annotation tool Help dialog's browser-compatibility info.
- Update the production server sometime tomorrow evening.

StephenChan:
> - Confirm whether the Chromium-specific orientation problem is fixed or not.

This is indeed fixed in 4636c2d. It was the same problem that popped up in Chrome in April 2020. I just lost track of the timeline and thought this problem happened earlier than that, for some reason. I've done some testing with older browser versions to clear up the timeline of events, which I'll re-summarize in further issue comments later.

> - Update the annotation tool Help dialog's browser-compatibility info.

Just added a commit for this.

@beijbom Let me know if you get a chance to look over the new commits today, but if not, I'll still update production this evening.

beijbom:
@StephenChan : new changes look good. Nice work on this one!

StephenChan:
Thanks. Server updated; will also work on a potential blog post.

Issue #332, opened 2020-06-14
Problems with EXIF-rotated thumbnails


StephenChan:
Branched off from #107. Our generated thumbnails take EXIF orientation into account, in the sense that easy-thumbnails rotates the pixel data and then strips the EXIF tag when generating the thumbnail.

It's been this way for a long time. easy-thumbnails had thumbnails using EXIF orientation all the way back in [2011/12/05](https://github.com/SmileyChris/easy-thumbnails/commit/cfb6b8b2ae49f2ff2bad19ab20ee793cc755e3c3)! That was before their version 1.0 release in mid-2012.

However, we actually had easy-thumbnails installed previously as an alpha version, since it was a [django-userena](https://github.com/bread-and-pepper/django-userena/blob/1601c2ab7c573ebb403ecf8e51c978a6360f759d/setup.py) requirement. [Here's a 2011/08/06 commit](https://github.com/beijbom/coralnet/commit/43d084f25b32c67d02577a8e84be2156dc587664) of ours using easy-thumbnails. I'm not sure what our easy-thumbnails update history looks like, unfortunately. It's actually plausible that we stuck with our mid-2011 installed alpha version for quite a while.

Anyway, this behavior in itself isn't necessarily a bad thing, but there are a few problems related to the situation:

## Problems

1. The annotation area edit UI uses a thumbnail, so EXIF orientation is applied if it's present. However, the edit UI doesn't take EXIF orientation into account. So the thumbnail gets stretched to fill the space that the un-rotated image would have filled.

   <img src="https://user-images.githubusercontent.com/857711/84535330-796aab00-aca0-11ea-8963-30797558ea85.jpg" width="400">
   
2. For a long time, easy-thumbnails didn't handle EXIF orientation correctly in the 90 degree rotation case. Basically, the image content gets rotated in the correct direction, but the content gets saved into the unrotated aspect ratio. So part of the image gets cut off, and it ends up with black bars.

   Example of an original image, displayed here WITHOUT its 90-degree EXIF rotation:

   ![thumbnail-black-bars_original](https://user-images.githubusercontent.com/857711/84584480-d5752280-adb9-11ea-9098-89635fecabb8.jpg)

   And its thumbnail:

   ![thumbnail-black-bars_thumbnail](https://user-images.githubusercontent.com/857711/84584481-d908a980-adb9-11ea-9763-1c723340c3d4.jpg)

[...]

PR #338, opened 2020-06-30
Thumbnails: rotation fix, and better file reuse

StephenChan:
I figured it'd be best to address issue #332 before advising people to double check a bunch of images, because when you load the annotation tool, it generally takes several seconds for the thumbnail to get replaced by the full image. If the thumbnail is the wrong orientation, and the user is a fast checker, then they might check the points against the wrong image orientation.

While I was at it, I consolidated some very similar thumbnail sizes (image detail, annotation tool, annotation area edit) to reduce thumbnail storage volume and generation time.

### DIFF
--- a/project/annotations/templates/annotations/annotation_area_edit.html
+++ b/project/annotations/templates/annotations/annotation_area_edit.html
@@ -32,8 +32,17 @@
     </form>
 
     <div id="image_container">
-        <img id="image" src="{% thumbnail image.original_file thumbnail_dimensions %}" />
-        <div id="annoarea_box"></div>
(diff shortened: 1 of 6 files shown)


beijbom:
Looks good. So this mean you will also purge existing thumbnails?

StephenChan:
> Looks good. So this mean you will also purge existing thumbnails?

Yeah - existing thumbnails for images with non-default EXIF orientation. Code to purge them is here: https://github.com/beijbom/coralnet/issues/332#issuecomment-651478211 I'll start running that tonight; it'll take several days to complete.

PR #344, opened 2020-08-19
Increased delay for uploads

beijbom:
Addresses https://github.com/beijbom/coralnet-system/issues/31.

This hasn't really been tested, but I figured let's try it and see if it solves the problem.


beijbom:
> Cool. It might be nice to leave an explanatory comment in the code there, but up to you.

Good idea: added.

Issue #348, opened 2020-10-04
Browse Patches: annotator dropdown is slow to populate


StephenChan:
https://github.com/beijbom/coralnet-system/issues/3 has a lot of performance related things going on. I'd just like to put this particular problem in its own (public) issue for future reference, even if I already have a PR lined up.
            
In some sources with many annotations, like source 1052, arriving on the Browse Patches page (before doing a search) takes 30-90 seconds, while Browse Images doesn't have the same slowness. This seems to be caused by the 'Annotator' dropdown field being populated by an inefficient DB query.
            
If issue #347 is implemented before this is optimized, the slowness would also affect Browse Images.

beijbom:
On Sun, Oct 4, 2020 at 16:31 StephenChan <notifications@github.com> wrote:

>
>
> beijbom/coralnet-system#3
> <https://github.com/beijbom/coralnet-system/issues/3> has a lot of
> performance related things going on. I'd just like to put this particular
> problem in its own (public) issue for future reference, even if I already
> have a PR lined up.
>
>
> In some sources with many annotations, like source 1052, arriving on the
> Browse Patches page (before doing a search) takes 30-90 seconds, while
> Browse Images doesn't have the same slowness. This seems to be caused by
> the 'Annotator' dropdown field being populated by an inefficient DB query.
>
>
> If issue #347 <https://github.com/beijbom/coralnet/issues/347> is
> implemented before this is optimized, the slowness would also affect Browse
> Images.
>
> If I understand you correctly, #349 fixes both throw and #347, so the new
slowness for browse images would be fixed immediately, right?

StephenChan:
That's right.

StephenChan:
Addressed in PR #349, but I'd rather keep this open so that I remember to double check the performance with source 1052 specifically, once this is actually rolled out in production.

StephenChan:
> I'd rather keep this open so that I remember to double check the performance with source 1052 specifically, once this is actually rolled out in production.

Hmm. First patches load took like 60 seconds, then 10-15 seconds after that. So, maybe a little faster, but the performance here is still worth another look.

Issue #347, opened 2020-10-04
Add 'last annotation' image filters


StephenChan:
Issue #330 has a few things going on, so I wanted to separate this part into its own issue (even if I already have a PR lined up).

- In Browse Images and Edit Metadata, be able to filter by the date of the last annotation activity on the image, and the method/user responsible for the last annotation activity.

- While we're at it, add an annotation date filter to Browse Patches as well.

Example applications in Browse Images:

- View images that were annotated recently

- Count how many images were annotated in, say, the last month

- View recent activity by a particular annotator; for example, one student of many working in a particular source

StephenChan:
Done in PR #349.

PR #323, opened 2020-05-25
Beta2 rollout

beijbom:
This is the base-branch for the new back-end rollout:

- [x] Spacer integration. Issue #264 
- [x] Switch to running jobs through AWS Batch [@beijbom ]
- [x] Add tables to track all backend jobs. Cut reliance of SQS to track completed jobs. [@beijbom]
- [x] Unskip currente backend tests. [@StephenChan ]
- [x] More vision backend tests. [@StephenChan ]
- [x] Support for different extractors. Issue #235  [@StephenChan ]

BONUS (can be pushed to later)
- [ ] More vision backend tests & SWE [@beijbom ]
- [ ] Purge python 2.7 compatibility code. Issue #364 [@StephenChan ]
- [ ] Make reg-tests robust for all settings. Add comparison for classifier performance. [@beijbom]

Issue #46, opened 2016-09-05
Add sort-method field on Browse Search form


StephenChan:
Currently, images are always sorted by filename, and patches are always shuffled randomly. The sort-method field would give more flexibility.

This field can be part of the search form. On Browse Images/Edit Metadata it can default to sorted by filename, with random being an option. We might consider allowing sorting by metadata fields as well, which could be useful for Edit Metadata in particular. Could also sort by upload date.

On Browse Patches it would be either sorted by machine confidence, or shuffled randomly. Another idea is sorting by annotation date.

PR #349, opened 2020-10-04
Last annotation field + sorting options in Browse

StephenChan:
Addresses the following:

- Issue #347 (last-annotation filters)
- Issue #348 (Browse Patches performance)
- Part of Issue #46 (sorting options); only implemented for the Browse Images page and the actions available from there. Sorting options are: image name, upload date, photo date (in metadata), last annotation date.

Regarding release timing: since we just released the blog post talking about the API, I think I might as well wait a bit before actually rolling this PR out in production (which would come with another announcement). It'll be good to merge this now though, since it's a decent chunk of code. After merging, I'll work on the vision backend stuff (#323). Then sometime later I'll roll out this PR.

There are two added migration files, one for adding the field itself, and one for populating it. The data migration doesn't work on Python 3. It seems to fail when trying to instantiate the Image model class, but I wasn't able to determine the root cause. So that's annoying, but it isn't going to block progress:

- Our master branch still works on Python 2, so we can run the migration in Python 2 for our production and development databases.

  If we have a future migration where we also have to instantiate Images, that may or may not be a problem when we're on Python 3 only. But I guess we can cross that bridge when we get there.

- When the migration is run as a test-runner initialization step, there are no images in the DB yet. So no Images get instantiated, and the error doesn't happen.
  
- For the migration test, I've added a `@skipIf` decorator which skips the test on Python 3.

As for the unit tests, I think they ended up being reasonably thorough. Notes:

- It might be nice to add a couple more tests to Metadata, Export, and Delete to make sure those use the new filters in the same way as Browse Images. They just use the same image-filtering form as Browse, though, so it doesn't seem like a likely issue. In any case, the already existing tests have been fixed as needed.

- I left a TODO comment in `test_annotation_tool.py` for a text check that I commented out. It's something which has undefined order now, but I anticipate that it may be consistent once we're on Python 3.6+ only, where dicts are ordered by default.

Some implementation details:

- I originally thought I could update `last_annotation` with an Annotation post-save signal without changing any other code. However, I was mistaken, because sometimes the code would look something like this:
  
  ```python
  image = Image.objects.get(pk=image_id)
  # <update annotations here>
  image.confirmed = True
  image.save()
  ```
  
  Updating annotations would update the Image's `last_annotation` via a signal. The problem is the `image` object in this function would still have the old `last_annotation` value. So, the `image.save()` statement would re-save the OLD `last_annotation` value.
  
  One solution is to always call `image.refresh_from_db()` after annotations are created/updated/deleted. So, that's kind of what I ended up doing - except I figured: if I already have to call `refresh_from_db()` anyway, why not call a function that updates `last_annotation`, refreshes `image` as needed, AND does other stuff like updating the `confirmed` flag?
  
  We already had a utility function `after_saving_annotations()` which was halfway to this idea, so I added onto that function and made sure it was used more consistently. It's now named `after_saving_points_or_annotations()`. With this solution, I ended up not needing to add a signal handler.

- Sorting was more complicated than I gave it credit for:

  - When figuring out the next or previous ordered image, it helps to know how a queryset has been ordered. It's possible to get this from a queryset's properties, but I think the properties I used are technically Django's internal API. So it's something to keep an eye on when updating Django versions. The relevant function is `_get_next_images_queryset()` in `images/utils.py`. The alternative was to pass an ordering parameter through a bunch of functions, which I didn't feel was a better solution.

  - Multiple images can have the same photo date (year/month/day), so I used the image primary key as a tiebreaker to ensure we have a consistent ordering.
  
  - Photo date and last annotation date can be null. PostgreSQL orders null after non-null values, so having well-defined behavior isn't an issue. But Django querysets don't accept null-comparison lookups such as `photo_date__gt=None`. So when figuring out what the next or previous ordered image is, if the 'current' image has a null date, I had to make special cases for that. Again, that's in `_get_next_images_queryset()`.


project/annotations/utils.py
```diff
@@ -163,14 +161,24 @@ def apply_alleviate(image_id, label_scores_all_points):
...
     if alleviate_was_applied:
-        after_saving_annotations(img)
+        after_saving_points_or_annotations(img)
 
 
-def after_saving_annotations(image):
+def after_saving_points_or_annotations(image):
     """
-    Run this function after saving a batch of Annotations for an Image.
+    Run this function after creating, updating, or deleting Annotations for
```
beijbom, line 169: I do wish we could get rid of this. It's a tall order for a new developer to know to call this method. In a perfect world we'd add a post-save hook to the Annotation model to execute these operations on the relevant image. Did you look into that option? Or perhaps something to consider for next round?

beijbom: Ok, so I looked a bit more at your comment and at this code. It seems to me we could add this as a post-save hook on `image.save()`.  We could also add a smaller post-save hook to `annotations.save()` which does this part:
```
last_annotation = image.annotation_set.order_by('-annotation_date').first()
image.last_annotation = last_annotation
image.save()
```

We'd also leave the code in the image save post-hook, so the use-case you discuss:
```
image = Image.objects.get(pk=image_id)
# <update annotations here>
image.confirmed = True
image.save()
```
should be fine.

beijbom:
@StephenChan : nice work on this. Migrations and tests ran fine on my dev-system. I don't like that python3 thing, but I agree with your work-around and we can indeed worry about it later. I added a suggestion on a post-hook. Let me know what you think.

project/annotations/utils.py
StephenChan: Hmm, but if the Annotation post-save hook calls `image.save()`, then that means the Image post-save hook also runs as a result, right?

Looking at all the places where this PR calls `after_saving_points_or_annotations()`, I think some kind of post-save signal (or an overridden save() method, come to think of it) would work fine in most places. There are still two cases I'm concerned about though:

1. During the `annotation_tool` view, after applying Alleviate, the Image object is then passed to the `annotation_tool.html` template, where its last_annotation value gets read. If the last_annotation gets updated by a signal / custom save method on Annotation (instead of by an 'after saving' function call), then the template still reads the old value of last_annotation. Unless we remember to call `refresh_from_db()` on the Image.

   I agree that having to remember the 'after saving' call isn't great, but I think having to remember a 'refresh' call is probably harder than that.

2. During the `upload_annotations_ajax` view, we save some annotations, and then still have some Image fields to update after (point generation method, CPC content/filename). So there's still a danger of overwriting the old value of last_annotation, unless the annotation saving code is moved later, *after* all the other Image fields are updated. But putting the updates in the proper order seems like another thing that's more subtle to remember compared to an 'after saving' call.

StephenChan:
> I clicked around a bit on my dev-system, but I wasn't sure how to test the new functionality.

Go to Browse Images, it should have the new fields at the bottom of the search box.

1_browse-images

project/annotations/utils.py
beijbom: > Hmm, but if the Annotation post-save hook calls image.save(), then that means the Image post-save hook also runs as a result, right?

Good point! So all we need to do in the annotation post-save is calling `image.save()`. That's even easier. 

For you new cases, let's see.

1) In this case, the extent of the damage is that we show a stale value of `last_annotation`? That seems fine to me. Or are there more severe implications? Of course, we could also do a `refresh_from_db()` call to refresh the item in memory. The reason I'm advocating for making the "after saving" call a hook is that it updates the DB state, which is critical to keep updated.

2) I don't understand the issue here. In my suggestion the `image.save()` and `annotations.save()` post hooks would both update last_annotations field. So it shouldn't matter which order you do the saves. Can you write out the sequence of events you are worried about?

StephenChan: > 1. In this case, the extent of the damage is that we show a stale value of `last_annotation`? That seems fine to me. Or are there more severe implications? Of course, we could also do a `refresh_from_db()` call to refresh the item in memory. The reason I'm advocating for making the "after saving" call a hook is that it updates the DB state, which is critical to keep updated.

It shows a stale value on the annotation tool page in the case that Alleviate was applied. We already don't make it very clear to the user when Alleviate was applied, so I think the stale value can be quite misleading.

But yes, it's not as critical as keeping the DB state updated. And the annotation tool IS a rather weird case, because it might be the only data-saving view which is a GET instead of a POST, and doesn't just redirect to a 'read-only' view after it's done saving. So I'd be okay with adding more weirdness with a `refresh_from_db()` if it's just this case. (I've been thinking annotation tool should be split into separate POST/GET views, which would make this moot, but I guess that's another task for later.)

> 2. I don't understand the issue here. In my suggestion the `image.save()` and `annotations.save()` post hooks would both update last_annotations field. So it shouldn't matter which order you do the saves. Can you write out the sequence of events you are worried about?

Oh, my bad. Yeah, I think you're right about this, and my brain just hadn't caught up to the idea yet of `image.save()` having a hook.

I guess the only other concern I can think of is performance. I can try saving 200 annotations in the annotation tool, thus calling the hook 200 times, and see how it compares to the old code.

beijbom:
Sounds good. Lmk how those test go.

StephenChan:
Tests on the staging server:

- Saving 200 annotations with the old code: 4 seconds
- Saving 200 annotations using save hooks (see new commit): 5-6 seconds

So it's a little slower, but acceptably so, IMO.

I ended up using custom model-save() methods instead of post-save signals, because calling `image.save()` within an Image post-save signal handler would result in infinite recursion. Also, I needed hooks on Annotation deletion and Point saving as well.

I do like this scheme now, since it almost completely isolates concerns of the two 'redundant' Image fields (`confirmed` and `last_annotation`) to the model-save() methods. The only exceptions are the 'improper' views which do both saving and reading of data (annotation tool's main view, as I mentioned previously, and also `save_annotations_ajax` which returns the 'all done' boolean value).

One other thought on annotation-tool save performance: resolving issue #32 could significantly boost performance, if we could implement it in such a way that saves Annotations in bulk, rather than one by one.

project/annotations/models.py
```diff
@@ -32,6 +32,16 @@ def label_code(self):
...
 
+    def save(self, *args, **kwargs):
+        super(Annotation, self).save(*args, **kwargs)
+        # The image's annotation status may need updating.
+        self.image.save()
+
+    def delete(self, *args, **kwargs):
+        super(Annotation, self).delete(*args, **kwargs)
+        # The image's annotation status may need updating.
+        self.image.save()
```
beijbom, line 43: Did you test this? Can you access `self` after you deleted it?

beijbom (approved):
I like this version much better. Nice!

project/annotations/models.py
StephenChan: Good question - that unintuitive-ness hadn't occurred to me. But it does indeed work. After the `super` call, `self.id` is None, but `self.image` is still available. It's tested in `images.tests.test_image_detail_actions.DeleteAnnotationsTest.test_last_annotation_field_updated`.

beijbom:
Gotit. Ok!

beijbom:
@StephenChan : I just tuned the CI pipeline to remove the block (it was a leftover from when I switched to travis.com from travis.org. This is ready to merge.

StephenChan:
@beijbom Ah, I was wondering about that. Thanks!

PR #351, opened 2020-10-19
Annotation tool: Add another EXIF-loading error case

StephenChan:
On the production site, if you visit image ID 1508487 in the annotation tool, you get this Javascript alert message:

> Error when loading the image: "Exif might be wrong. Got incorrect value type to decode. type:0" If the problem persists, please contact the admins.

This PR should fix that. We don't have automated tests for Javascript yet, though. To manually test, download image 1508487 from the production site, upload it to your development site, visit it in the annotation tool, and see if you get the above Javascript alert or not.


beijbom (approved):

beijbom:
Didn't test in develop, but looks good to me.

PR #355, opened 2020-10-22
Aws batch integration

beijbom:
## Changes

- Changes to `pyspacer == 0.2.9` which supports AWS Batch. 
- Implements a BatchQueue for handing submission and collection of jobs.
- Adds a BatchJob table to the `vision_backend` module. 
- Remove dependencies on SQS so I removed all those settings.

## How to check
Run
```
python manage.py vb_regtests
```
Then monitor
http://127.0.0.1:8000/admin/vision_backend/batchjob/
and 
https://us-west-2.console.aws.amazon.com/batch/old/home?region=us-west-2#/dashboard


project/config/settings/production.py
```diff
@@ -35,8 +35,7 @@
 SPACER_QUEUE_CHOICE = 'vision_backend.queues.SQSQueue'
```
StephenChan, line 35: This should become `BatchQueue`, right?

StephenChan:
That was fast! Nice job. The new table looks good, and the regtest worked fine for me, including monitoring using the links you provided.

I see there are still a couple of references to SQS in the code, including the production settings. Did you just miss those, or is there a reason for leaving them there?

beijbom:
> That was fast! Nice job. The new table looks good, and the regtest worked fine for me, including monitoring using the links you provided.

Thanks for quick review.  Glad it worked.

> I see there are still a couple of references to SQS in the code, including the production settings. Did you just miss those, or is there a reason for leaving them there?

I changed the production settings -- thanks for catching that! The other mentions are in the SQSQueue class itself and the factory. I think we can leave those in since the class itself is correct. I did add a ValueError if someone tries to initialize a SQSQueue through the factory.  LMK what you think.

StephenChan:
> I changed the production settings -- thanks for catching that! The other mentions are in the SQSQueue class itself and the factory. I think we can leave those in since the class itself is correct. I did add a ValueError if someone tries to initialize a SQSQueue through the factory. LMK what you think.

Okay, yeah, that works. We can always remove SQSQueue sometime later if we're sure we don't need any part of it anymore.
Looks good to merge then.

PR #356, opened 2020-11-08
api deploy tests unskip

beijbom:
…h LocalQueue it gets `TypeError: Object of type 'int32' is not JSON serializable`.

### DIFF
--- a/project/config/settings/dev_beijbom.py
+++ b/project/config/settings/dev_beijbom.py
@@ -2,8 +2,8 @@ from .base_devserver import *
 
 ALLOWED_HOSTS = ['127.0.0.1', 'testserver', 'localhost']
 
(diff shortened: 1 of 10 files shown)


StephenChan:
@beijbom That particular test (`test_deploy.SuccessTest.test_done`) should be unskipped now. I'll work on unskipping the other 3 deploy-related tests tomorrow.

StephenChan:
@beijbom That's all the deploy tests unskipped, so if this looks good to you, perhaps we can merge it into `beta2-rollout`. Then I'll work on another PR for adding more tests. (Dupe rows/cols, `clean_up_old_batch_jobs`, and then anything else I can think of as time allows.)

beijbom:
Sounds great and this looks good to me. Thanks for pushing this one out. I'll merge this to the feature branch now.

PR #357, opened 2020-12-01
More tests for vision backend tasks

StephenChan:
For #323.

- Tests various additional cases in feature extraction, training, and classification. The tests for these tasks are now split up into `vision_backend/tests/tasks/test_extract_features.py`, `test_train.py`, and `test_classify.py` respectively.

  - I removed the class `ClassifyUtilsTest` because `test_train.py` now has equivalent tests for everything that used to be in that class - except now, the classify task is actually called, instead of being mostly-duplicated in the test code.

- Tests duplicate rowcols in feature extraction, training, and classification. See the `test_with_dupe_points` tests in the above-mentioned three files. @beijbom - two questions about this:

  - Just to confirm, if an image has 10 points with one duplicate among them, then here's what currently happens: we get 9 features, and we get 10 data points for traindata / valdata / valresults. Does this sound like what it should be?

  - The `assert len(image_labels) == len(image_features.point_features)` statement in spacer's `train_utils.py` causes the `test_train.py` dupe-points test to fail. When I tried commenting this assert out, the test passed. What would the best course of action be here? Remove the assert, modify the assert to accommodate dupes, or something else?

- Tests the `clean_up_old_batch_jobs` task. See `vision_backend/tests/tasks/test_misc.py`, which now also contains the other tests which used to be located at `vision_backend/tests/test_tasks.py`.

- (Slightly off-topic, but:) Makes `vision_backend_api.tests.DeployResultEndpointTest.test_success` more robust by not assuming certain scores. Previously, the test had failed a couple of times for me due to varying scores returned by deploy.

I wasn't super thorough, but overall I added a decent number of test cases for the three main tasks (extract, train, classify) and was able to test everything I tried to test. So I think vision backend testability has been greatly improved, which was one of our main goals with all this!

One design comment I have - it's slightly awkward to have two minimum train-image thresholds, one in CoralNet (`MIN_NBR_ANNOTATED_IMAGES`), and one in spacer (`MIN_TRAINIMAGES`). I think ideally, spacer would provide a way to override the default `MIN_TRAINIMAGES` value, though I'm not sure what the best way would be off the top of my head. Not too big of a priority though; the biggest benefit for now would be speeding up VB tests by being able to set a lower `MIN_TRAINIMAGES`.

Anyway, the main issue to resolve for now is the spacer assert breaking the `test_train.py` dupe-points test.


StephenChan:
Ah, Travis gets a second fail which didn't happen for me locally - `test_classify_unconfirmed_image (vision_backend.tests.tasks.test_classify.ClassifyImageTest)`... I'll look into that.

StephenChan:
Okay, I didn't really understand that second fail, but it passes for now at least. Just the dupe points thing to resolve then.

beijbom:
@StephenChan : Thanks for finding that error. There were 2 things going on.
1) That assert on spacer side was a bit strict. I relaxed it in https://github.com/beijbom/pyspacer/pull/37.
2) However, the assert can only be relaxed for features stored using the new ImageLabels data-class. For legacy we need to still compare the number of row, col locations. So I changed this line: https://github.com/beijbom/coralnet/pull/357/commits/592f2ee3c4e9d6c105ea9556baf429e92e79be69 and now it works. Since we did change that we don't need the more relaxed assert, so we don't need to bump the pyspacer version.

StephenChan:
Thanks @beijbom! That all makes sense, and your changes look good.
Let me know if you have comments about the rest of the PR, else I'll just merge this next week when needed.

beijbom (approved):
Looks great! Thanks for adding all of these.

Issue #235, opened 2019-08-16
Enable support for different feature extractors.


beijbom:
- Add string field to Source model to indicate which version of the feature extractor the source uses. 
- For all existing sources set the legacy extractor.
- Set the new feature extractor as default for new sources. Users can change this to the old extractor during source creation.
- Allow users to change extractor in the source admin page. 
- If the feature extractor is changed, warn user and wipe all features and classifiers from source.
- Add valid choices for feature_extractor_names to vision_backend settings.

StephenChan:
1) Just to confirm the relative priorities of issues here - the new spacer rollout would happen before the new feature extractor rollout, right?

2) On this:

   > - Allow users to change extractor in the source admin page.

   Although maybe not necessary, it's worth also considering issue #92 on how to present the options. We would now have two backend-related settings, one for confidence threshold and one for feature extractor. The thing is, both require some explanatory text, so I don't think they'll be too lonely on a single 'backend settings' page.

3) Note from email discussion on how the code would change:

   > The extract_feature message has a string "feature_extractor_name". I'm planning to read that out from the source. So for this ticket we just need to add a string choice field to the source and add the valid choices to settings.

beijbom:
Regarding priorities. The reality is that we will have a pretty large feature-branch loop unfortunately. Basically, we will need to roll-out python3 and the new spacer at once. This means that once we cut the python3 branch for
https://github.com/beijbom/coralnet/issues/59, we need to complete this ticket, along with #264 and #302 into that branch before we can merge to master an update the server. We should also plan a few days of downtime when we do that so that we can do proper shake-outs.

StephenChan:
Huh. Well, that's going to be fun then. Do you think we could ease things a bit by implementing some of this ticket in advance - like, implementing the extractor-choice field, but with only one option instead of two?

Also, regarding #59, I could aim to master-merge as many 'safe' Py3 porting changes as possible, before really carving out that final Py3 branch which would be merged along with spacer. Ideally, that earlier merging step would include porting everything that's not too closely tied to the vision backend.

StephenChan:
Implemented in PR #360.

PR #360, opened 2020-12-19
Enable support for different feature extractors

StephenChan:
For issue #235.

There are two new DB migrations. The first migration adds a feature extractor field to the Source model, with a default of `vgg16_coralnet_ver1`. This sets the value for all existing sources. The second migration changes the default to `efficientnet_b0_ver1`, establishing the default for sources created from here on.

The new/edit source forms have a field for the feature extractor (help text isn't there yet):

image

When editing a source, if the choice is different from before, this warning appears:

image

And then when you click 'Save Changes', you get a confirm message:

confirm

You can see the current setting on the source main page. The position of the "(edit)" link is kind of bugging me now, since it seems to imply that only the confidence (not the other stuff) can be edited with the link, but I left it for now:

image

@beijbom Two things: 1) What should go into the help text, or do you want to just write it yourself? 2) The extractor choices are currently displayed as "EfficientNet (default)" and "VGG16 (legacy)". Is this good, or is it better to include 'B0' or 'ver1' in the displays?


project/images/templates/images/form_source.html
```diff
@@ -77,6 +77,25 @@
...
     {% endif %}
+
+    {% if source_form_type == "new" or source.enable_robot_classifier %}
+        <hr class="light"/>
+
+        <legend class="smaller">Feature extractor</legend>
+
+        <span class="helptext_long">
+        {# TODO #}
+        (TODO: Explain what this is and how to choose it.)
```
beijbom, line 88: Suggested text:

We recommend EfficientNet extractor for all use-cases. It is faster and 2-3% more accurate on average. It is a more modern neural network architecture and trained on more data. The legacy VGG16 extractor is provided only for sources that want to retain their old classifiers, for example, if they are already deployed in a survey.

StephenChan: Sounds good, added a commit for that.

project/annotations/management/commands/replace_label_in_source.py
```diff
@@ -102,5 +102,5 @@ def handle(self, *args, **options):
         source.labelset.locallabel_set.get(global_label=old_label).delete()
 
         # Reset the backend state for this source.
-        backend_tasks.reset_after_labelset_change.apply_async(
+        backend_tasks.reset_backend_for_source.apply_async(
```
beijbom, line 105: This is actually not correct. `reset_after_labelset_change ` doesn't reset the feature extractions, only the classifications. So you need to maybe keep the old one and then create a new one `reset_backend_for_source` that calls `reset_after_labelset_change` but also `reset_features` for all images in the source.

beijbom: ... obviously, rename the function however you like. The old names are a bit weird.

StephenChan: Ah, thanks for the catch. Yeah, I definitely misread what that task was doing. I'll fix it later today.

StephenChan: Added a commit for that. Took longer than expected because I wanted to improve some related unit tests.

beijbom: Tnx. Let me try it out.

StephenChan: I just added 1 more commit. The reset-task tests now check that the backend objects get re-created after being deleted.

beijbom (approved):

PR #386, opened 2021-04-13
fixing logic for batch cleanup jobs

beijbom:


beijbom:
@StephenChan : I kept getting these error emails and I finally figured out what was wrong. This should be more correct, right? We just want to delete jobs that were too old and never actually SUCCEEDED? Alternatively, we can purge that table for any old jobs including those that SUCCEEDED but we shouldn't email the admins about that.

beijbom:
... Hmm. I'm still not sure if what I wrote is correct. Arguably, we want to mail admins if there are old jobs that never succeeded. But I'm not sure we want to delete the entries since that makes it very hard to debug what happened.

So my suggestion is this:
We delete all SUCCESSFUL jobs just to keep that table to a manageable size. 
We send out a warning email if there are jobs older than say 5 days. Those will likely never finish so we should understand why.

StephenChan:
Ah, well you're right that the old cleanup-task code isn't what we want. This task should be the only way that jobs are getting deleted from the database, so it seems to mail us about every single job, regardless of whether it completed or not. Replicated below for reference:

```python
@periodic_task(...)
def clean_up_old_batch_jobs():
    thirty_days_ago = timezone.now() - timedelta(days=30)
    for job in BatchJob.objects.filter(create_date__lt=thirty_days_ago):
        job.delete()
        mail_admins("Job {} not completed after 30 days.".format(
            job.batch_token), str(job))
```

I think the simplest fix would be to make the mail conditional on the job status. Since FAILED already sends an email in queues.py `collect_job()`, we can probably exclude both SUCCEEDED and FAILED here:

```python
@periodic_task(...)
def clean_up_old_batch_jobs():
    thirty_days_ago = timezone.now() - timedelta(days=30)
    for job in BatchJob.objects.filter(create_date__lt=thirty_days_ago):
        job.delete()
        if job.status not in ['SUCCEEDED', 'FAILED']:
            mail_admins("Job {} not completed after 30 days.".format(
                job.batch_token), str(job))
```

If we want to report stuck jobs after 5 days instead of 30 days, I guess we can have the warning email in a separate task. So we'd first simplify the cleanup task like this:

```python
@periodic_task(...)
def clean_up_old_batch_jobs():
    thirty_days_ago = timezone.now() - timedelta(days=30)
    for job in BatchJob.objects.filter(create_date__lt=thirty_days_ago):
        job.delete()
```

And then there'd be a few ways to write the 'warn about stuck jobs' task. The way you coded it in this PR would send one email per stuck job per day, but that might be excessive. I think there are two better options, either one email per stuck job only once, or one email per day when there are stuck jobs present:

```python
@periodic_task(...)
def warn_about_stuck_jobs():
    """
    Warn about each stuck job that has crossed the 5 day mark. Since this
    task runs every day, only grab jobs between 5-6 days so we don't issue
    a repeat warning after 6 days, 7 days, etc.
    """
    five_days_ago = timezone.now() - timedelta(days=5)
    six_days_ago = five_days_ago - timedelta(days=1)
    stuck_jobs_between_5_and_6_days_old = BatchJob.objects \
        .filter(create_date__lt=five_days_ago, create_date__gt=six_days_ago) \
        .exclude(status='SUCCEEDED') \
        .exclude(status='FAILED')
    for job in stuck_jobs_between_5_and_6_days_old:
        mail_admins("Job {} not completed after 5 days.".format(
            job.batch_token), str(job))
```

```python
@periodic_task(...)
def warn_about_stuck_jobs():
    """
    Warn if there are jobs that have been stuck for over 5 days.
    """
    five_days_ago = timezone.now() - timedelta(days=5)
    stuck_jobs = BatchJob.objects \
        .filter(create_date__lt=five_days_ago) \
        .exclude(status='SUCCEEDED') \
        .exclude(status='FAILED')
    if stuck_jobs.exists():
        example_job = stuck_jobs[0]
        mail_admins("Job {} and {} others not completed after 5 days.".format(
            example_job.batch_token, stuck_jobs.count() - 1), str(example_job))
```

What do you think?

beijbom:
> If we want to report stuck jobs after 5 days instead of 30 days

Yeah, I think we do. Or at least we don't want to delete the failed jobs entries in case we need them for debugging purposes. 

As for the clean-up of old jobs in the first place. Do we want / need that? Why are we worrying about cleaning up this table when there are other much larger tables?

I like the first version of the "warn about stuck jobs" task.

StephenChan:
> As for the clean-up of old jobs in the first place. Do we want / need that? Why are we worrying about cleaning up this table when there are other much larger tables?

Well, I think keeping the old BatchJobs would be fine. I think we were cleaning up old ApiJobs so that the API-job monitor page wouldn't get absurdly long. And then one of us copied over the cleanup task to BatchJobs, but I don't remember if there was a reason we had to.

beijbom:
Ok, cool. I updated the code. I ended up keeping the after-30-day-cleanup. Feel free to remove it if you like.

StephenChan:
Task changes look good. For the unit tests, better to add the `BatchJobCleanupTest` back in, and write a new test for the `warn_about_stuck_jobs` task. I can do that tomorrow.

StephenChan:
@beijbom OK, added tests.

StephenChan (approved):

beijbom:
Thanks @StephenChan . Can you ping me once this is deployed to prod? I'd like to work my way through those errors and see if  they can be fixed properly by re-extracting the features.

StephenChan:
Sure, will do.

StephenChan:
@beijbom It's in production now.

PR #396, opened 2021-04-25
removing outdated messages in queue stats from vision overview page

beijbom:
We can add the batch queue back in later. This will at least bring the page back up.


StephenChan (approved):
Works for me (both with and without the 3rd commit you just pushed).

StephenChan:
@beijbom I was thinking of updating the production server this weekend, but let me know if you'd like this fixed sooner than that.

StephenChan:
@beijbom Production's updated now.

Issue #394, opened 2021-04-25
Update Pillow to 6 or later


StephenChan:
Not urgent currently, but would be nice to get out of the way. May not be super easy to do though. I gave this a quick try and got: `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process` on a bunch of tests.

beijbom:
Interesting. I know windows can be a bit fickle with these sorts of things.
Tests are running fine in CI: https://github.com/beijbom/coralnet/pull/398.

PIL is only imported in 5 places. Can you see if you can resolve the issues
on your local dev system? This page has some hints:
https://stackoverflow.com/questions/27215462/permissionerror-winerror-32-the-process-cannot-access-the-file-because-it-is

StephenChan:
I'll look into it now.

StephenChan:
Completed by PR #398.

PR #398, opened 2021-04-25
bumping pillow

beijbom:


beijbom:
@StephenChan : tests run fine on my system and in CI.

beijbom:
I didn't remove `piexif==1.1.3` since it required some code-changes; we can do that in subsequent PR.

StephenChan:
I resolved that Windows issue mentioned in #394. It turned out to be caused by the code in only one test, even though I saw a bunch of tests fail. It was the test's file-cleanup routine that got an error, and each subsequent test tries to use the same directory for storing test files. So once that one test failed to clean up, every test running after it also failed.

(Admittedly, I don't exactly understand why my fix worked, but I suppose it did...)

StephenChan (approved):

PR #399, opened 2021-04-25
adding basic linting; updating documentation

beijbom:


beijbom:
@StephenChan : this is a rather light linting config I have used elsewhere. It will change for unused imports, blank lines, extra spaces and other basic syntax. It will fix the most obvious ones immediately, for other it will just want you and prevent the commit from going through. LMK what you think.

beijbom:
I'll merge this. If you prefer not to use this, then you can simply not install the pre-commit hooks. Or feel free to change the settings in subsequent PRs.

StephenChan:
All right. Seems like a good idea, I'll try it out when I get the chance.

Issue #154, opened 2018-06-03
Label list and label detail pages are still somewhat slow to load


StephenChan:
Experiment more with pre-computing and/or caching here. We're seeing maybe 40 seconds for the label list if viewing for the first time in a day, and 10-15 seconds for Turf's label detail page, for example.

StephenChan:
PR #403 computes the label popularities in an asynchronous task, and also reduces the label list's DB query count related to functional groups. When the changes are pushed to the production website, we should check whether the performance is good enough now.

StephenChan:
As far as I can tell, PR #534 improved this to a point where it's no longer one of the bigger concerns with site performance.

PR #403, opened 2021-06-07
Add default calcification rates to labels + label list performance

StephenChan:
For issues #368 and #154.

- Added a Django management command to import default calcification rates from a CSV file.

  Since this is a Django management command, these CSV files can only be uploaded by someone with direct server access (currently Oscar or me). So Travis could send me the CSV file for me to upload.
  
  Example command usage:
  
  `python manage.py import_default_calcify_table calc_rates_20210518.csv "Default rates 2021-05-18" "Description goes here"`
  
  The command takes 3 parameters:
  
  - The CSV file path.
  - Name to save for the table(s)
  - A description to save for the table(s)

  Example of the CSV column format below. I stuck with using column titles like "Mean rate" instead of "mean_rate" because that was more consistent with the column titles in other CoralNet import functions, such as importing annotations or labelsets:

  Label | Region | Mean rate | Lower bound | Upper bound
  ----- | ------ | --------- | ----------- | -----------
  Acropora | Atlantic | 19.21 | 14.34 | 24.31
  Acropora cervicornis | Atlantic | 31.73 | 25.39 | 40.76
  Dipsastraea | Atlantic | 8.76 | 8.65 | 8.89
  Dipsastraea | Indo-Pacific | 0.98 | 0.81 | 1.21
  Cliona | Atlantic | -5.17 | -7.3 | -3.04
  CCA (crustose coralline algae) | Atlantic | 0.36 | 0.3 | 0.43
  
  Uploading the above data actually saves two calcification-rate tables into the database, one for each specified region (Atlantic and Indo-Pacific in this case).
  
  If the table name "Default rates 2021-05-18" was specified in the command, then the actual table names end up being "Default rates 2021-05-18 - Atlantic" and "Default rates 2021-05-18 - Indo-Pacific". This full name including the region must be at most 80 characters. The name is used in the CSV filename, when we provide table download links on the site (see label-detail pages, below).
  
  The description must be at most 500 characters. This isn't displayed on the site yet, but it's something we can make use of later.
  
- Default calcification rates are now shown on label-detail pages.

  

  Here's what it looks like if rates aren't available for a label.

  
  
  Clicking the '?' button gives CSV-download links for the default rate tables (again, one table per region). Later we can add a link to a Zenodo repo with the rate tables' change history, as proposed in [Travis's comment here](https://github.com/beijbom/coralnet/issues/368#issuecomment-821331207). I'm open to any suggestions on the wording of the sentences here.
  
  
  
- The label list now indicates availability of default calcification rates for each label.

  
  
  The 'C' icon is present if default rates are available for at least one region.

- The labelset page for adding/removing labels now indicates rate availability in the popup for each label.

  
  
  The indicator says whether rates are "Available" for at least one region. Otherwise it says "Not available".

- Label-list load times should be more consistently fast compared to before. See commits 766a47b and 85b8e61 for details. (I ended up working on this now, because loading calcification rates was an additional performance concern for this page if coded poorly).


StephenChan:
Update to the '?' button's help text on the label-detail page.

Issue #388, opened 2021-04-15
Save multiple calcification rate tables per source


StephenChan:
Enhancement of #377. Allow saving multiple tables per source (not just 1), up to a reasonable/practical limit (like 5). This could be useful if the source's data spans both the Indo-Pacific and Atlantic regions, or if specifying more regions is desired. However, this feature alone would not allow doing a source-wide export with table 1 applied to some images and table 2 applied to other images.

At export time, you can select which table to use from a list of saved tables. As mentioned in #377, each table would have a name, and this helps to distinguish the tables from each other. The table list needs a way to delete individual tables - perhaps a delete button next to each table, with an 'are you sure' prompt when you click the button. There should also be a CSV download button next to each table.

When you upload a table, this adds the table to the list of saved tables. However, if the source already has the max number of tables (e.g. 5), the upload is disallowed, and you're prompted to delete one of the previously saved tables first. (There are more complex UI flows we could do here, but this is simplest and pretty reasonable.)

If #376 also gets implemented, there has to be a rule for picking the table used in the incremental source-wide stats computations. One idea is to use the table that was most recently used for an export. However, this rule has to be made clear somewhere, and a source outsider (without Edit/Admin access to the source) should not be allowed to change the table used. When the table is changed, the source-wide stats have to be recomputed from scratch.

StephenChan:
Done in PR #404. (Note that #376 has not been done.)

Issue #377, opened 2021-04-09
Save user-uploaded label calcification rates per source


StephenChan:
At export time, when a custom rates table is uploaded, don't just use the table for that export; save the table for future usage in that source. Enhancement of #372, and likely required for #376.

The simplest thing to do is to only save one custom table: the last one used in that source. This seems like it would be sufficient for many cases.

It should be possible to give a name to the table, specifying a date or anything else one might want there. That way, it can be determined whether the table is out of date and/or suitable for that export operation. It should also be possible to re-download the saved table as CSV. If there is a table rename feature, it should come with a warning about traceability - exporting the same table at different times under different names can cause confusion.

One extra design detail here: permissions in the export form. The ability to upload and save a table should be limited to those with Edit or Admin access to the source. For anyone else (View access, or visitors to a public source), there would only be the option to use a previously uploaded table; if there is no such table, then they have to pick one of CoralNet's default tables to export instead.

StephenChan:
Done in PR #404.

Issue #373, opened 2021-04-09
Stats exports: include summary stats


StephenChan:
At the bottom of the exported CSV, add a single row with the following info:

- Image Covers, CSV: percent coverage of each label, taken over all the images in the export
- Calcification rates: Calcification rate averaged over all the images in the export. As well as the export-wide contributions from each label, if the relevant optional columns are enabled.
- Other export types: Not applicable (as far as I can think of)

I believe this has been requested a few times over the years for the Image Covers CSV export, and our response was that people can compute the summary stats themselves with basic knowledge of Excel formulas. Which is true, but it's not too difficult for us to include it in the CSV, either - and part of the calcification-rate plans include displaying the latest summary stats somewhere on the source (#375), which means we want to compute said summary stats anyway.

Two possible methods of averaging:

1. Compute stats for each image and then average over all images.
2. Compute stats for each point-annotation and then average over all point-annotations.

If each image in the source has an equal number of points, and each image's annotation area covers roughly the same area of the benthic zone, then either 1 or 2 should be valid. If the source is set up differently, it's possible that one averaging method could be more valid than the other, but I'm not sure if any other source setups are common on CoralNet. So I suppose we just pick one method of averaging and go with it (and update the help text to clarify how it's done).

traviscourtney:
I think it would be nice to have the exported summary statistics for the mean, lower, and upper calcification rates so the confidence bounds are included in any summary calcification rates so that there's easy user access to the calcification rates and uncertainties for their project.

StephenChan:
I didn't specify that clearly here, but you're right - mean, lower bound, and upper bound should be included in the summary. If there are columns for each of those 3 (as specified in #369), then the summary-stats row specified in this issue should be able to have numbers for all 3.

StephenChan:
Done in #404 (calcification) and #405 (image covers).

Issue #372, opened 2021-04-08
Calcification rate export: Support user-specified label rates


StephenChan:
Prerequisite: #369

When exporting calcification rates for a source, #369 should have given two options for specifying label calcification rates:

- Use the Indo-Pacific default rates
- Use the Atlantic default rates

And here we'll add a third option:

- Specify custom rates

When the third option is picked, a CSV-upload form field appears which lets you upload the rates as a CSV. The CSV format could match the table format shown in issue #368.

The text for each of the first two options (Indo-Pacific default, Atlantic default) should include a download link to a CSV of the default rates (for that source's labelset). So the recommended way to specify custom rates would be to download one of the default-rates CSV files first, then edit the CSV file locally to change the default rates to custom ones, then upload the edited CSV using the form field provided.

Then once you click Go on the export form, the calcification rate computations use the custom rates instead of default rates.

See #377 for saving custom rates tables (on a per-source basis) for subsequent exports.

traviscourtney:
I think it would be really great for the custom rate table to have a unique identifier, perhaps the name of the uploaded .csv or some user-defined name so that the ID for that rate table could be included in the calcification export in #369 to increase traceability for where the calcification rate data came from.

If possible, I think the #377 implementation would be really nice such that users could upload their custom rate table .csv to an image source and have it saved directly to that source with a distinct name. I suspect this would improve traceability and consistency especially for larger collaborative CoralNet projects with multiple users.

StephenChan:
Yes, I think for a start we can use the name of the uploaded .csv, and make it clear to the user that the filename will be used as the table ID for traceability. Then if #377 is implemented (I hope I get to it), we can make it a little nicer and have a text field to enter a name - that's a good point that #377 would be important for consistency between source collaborators.

StephenChan:
Done in PR #404.

PR #404, opened 2021-06-12
Per-image calcification rate exports + rate table uploads

StephenChan:
Addresses the following:

- #369 - export function for per-image rates
- Half of #373 - summary stats (added for calcification, not for percent cover yet)
- #372 - accept custom label rate tables
- #377 - save custom label rate tables for later use
- #388 - save multiple custom label rate tables per source

After doing the earlier tasks, #377 and #388 turned out to be reasonably straightforward on top of accepting custom tables in the first place, so I implemented those.

To fnd the export form, go to the bottom of Browse Images, then select "Export" and "Calcification Rates" in the dropdowns. (I'd still prefer to move all the export functions to their own page for easier discoverability, but at this point I'm not sure if that will happen by the end of June.)

01_export

The export form has this dropdown for selecting the rate table. Choices are the custom tables uploaded to the source, followed by CoralNet's default tables:

02_table-select-dropdown

"Click here to manage label-rate tables" brings up this popup of the available tables:

03_manage

Here, the "Delete" and "Upload a new table" buttons are only available if you have Edit or Admin permissions to the source.

For the default tables, there are two "Download CSV" buttons available. The first CSV is filtered down to only the entries in the source's labelset - so if there are 20 labels in the labelset, then the CSV has 20 rows (with rates of 0 to fill in gaps as necessary). The second CSV is the entire default rate table.

Click "Upload a new table" to bring up this form:

The name field is required and the description field is optional. Note that the name field doesn't need to have a date in it; I just wrote dates in my example tables' names above, to demonstrate that the name field can help with bookkeeping / traceability of the rates. The expected CSV layout is almost the same as the one we use for uploading default tables - there are `Label`, `Mean rate`, `Lower bound`, and `Upper bound` columns, but no `Region` column.

You can save up to 5 custom tables in a source. If you already have 5, you have to delete an existing table before uploading another.

Note that after uploading a table, there is no way to rename the table or edit its description/contents. So you'd have to delete and re-upload the table if you want to change any of those things. Should be a relatively minor inconvenience though.

Now for the results of the per-image rate export - without the optional columns, you get something like this:

05_stats-csv

And with the optional columns - not all columns are visible in this screenshot, but basically there are "M" (mean), "LB" (lower bound), and "UB" (upper bound) columns for each label:

06_stats-csv-with-extra-cols

The optional columns are off by default, since there can be a lot of them.

## Technical notes

`CalcifyTablesHelper.js` is my first real attempt at using modern Javascript (mostly meaning ECMAScript 6) in CoralNet. I previously was sticking to older style Javascript for consistency with the rest of the codebase, and thought it would be better to modernize all of the JS at once sometime (along with adding JS tests to make sure things don't break). But I'm currently not sure when that can happen, and I figured there was no harm in using modern JS when I have to add a new module anyway.

Also, working on this PR gave me a few ideas on improving code organization in the export and upload code. The first two of these are partially implemented:

- Use the `SourceCsvExportView` class to reuse code among the source-level CSV export views.

- Use the `CsvFileField` class instead of `CSVImportForm` for more of the CSV upload views. It's not that much of a difference, but I think "single form with CsvFileField plus other fields" would generally make for cleaner view code than "CSVImportForm plus a second form with other fields".

- Use Javascript class inheritance to improve JS code reuse among the various Upload pages.

- In general, upload and export logic specific to particular models should go in those models' respective apps. Only generic classes, functions, etc. not specific to a particular model should go in the `upload` and `export` apps. (Related: issue #74) I believe pyspacer is similar in how it organizes serialization code.

### DIFF
--- a/project/calcification/forms.py
+++ b/project/calcification/forms.py
@@ -1,0 +1,44 @@
+from django.forms import Form
+from django.forms.fields import ChoiceField, MultipleChoiceField
+from django.forms.widgets import CheckboxSelectMultiple
+
+from calcification.utils import get_default_calcify_tables
+
+
+def get_calcify_table_choices():
+    default_tables = get_default_calcify_tables()
+    choices = [
+        (table.pk, table.name)
+        for table in default_tables]
+
+    return choices
+
+
+class ExportCalcifyStatsForm(Form):
+    rate_table_id = ChoiceField(
+        label="Label rates to use",
+        choices=get_calcify_table_choices)
+
+    optional_columns = MultipleChoiceField(
+        widget=CheckboxSelectMultiple, required=False)
+
+    def __init__(self, source, *args, **kwargs):
+
+        super().__init__(*args, **kwargs)
+
+        # TODO: The best thing to do is to remove all possibility of
+        # initializing this form without a labelset, and then remove this
+        # conditional.
+        if source.labelset:
+            labelset_size = source.labelset.get_labels().count()
+        else:
+            labelset_size = 0
+
+        self.fields['optional_columns'].choices = (
+            ('per_label_mean',
+             "Per-label contributions to mean rate"
+             f" (adds {labelset_size} columns)"),
+            ('per_label_bounds',
+             "Per-label contributions to confidence bounds"
+             f" (adds {labelset_size * 2} columns)"),
+        )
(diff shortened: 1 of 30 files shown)


StephenChan:
Took this PR out of draft mode, and updated the original post with all the progress.

Issue #408, opened 2021-08-10
Training fails due to point mismatches and cannot retry


StephenChan:
We've been getting an AssertionError on [this part of the spacer code](https://github.com/beijbom/pyspacer/blob/7355e0f522e484ae435d1bf6183ba001a2dd88e0/spacer/train_utils.py#L145) when training some sources -
`assert rc_labels_set.issubset(rc_features_set)`

It's the same error as https://github.com/beijbom/pyspacer/issues/38 although that issue's discussion-thread ended up mainly being about other concerns.

This seems to be the easiest way to make it happen right now:

- Upload images
    - This queues feature extract tasks
- Collect jobs (feature extract)
- Upload points + annotations for those images, enough to start training.
    - This queues reset-features tasks for each image *with a 1 hour delay* - see [upload/views.py](https://github.com/beijbom/coralnet/blob/83e2a6bef160369034f4dea4ff90ebd3fe29d842/project/upload/views.py#L497)
    - This queues a classifier-training task *with a 10 second delay* - see [annotations/models.py](https://github.com/beijbom/coralnet/blob/15e5db13c739b2382290ee35467ce47c4dde36b2/project/annotations/models.py#L105)
    - The 1 hour delay was implemented in PR #344, to ensure the points + annotations transaction completes before features are reset. However, the delay also means the training task is very likely to run before some (or all) of the reset-features tasks are run.
- If spacer runs the training job before running the feature re-extract jobs of the training images, then the points are mismatched (new points from DB + old points from old features) and the AssertionError happens.

After this happens, the source isn't able to recover to a state where it can retrain (unless more images are annotated). The logic on CoralNet's side says the source doesn't need a new classifier. Here's the reason for that:

- Since a train task gets queued whenever one image becomes fully annotated, it's common for multiple train tasks to be queued for the same source one after the other.
- However, we don't want to create 1000 classifiers in a source that annotates 1000 images. So if we queue the train task repeatedly, we want most of those tasks to abort before they can submit spacer train jobs. To do this, we create the classifier during the task, and then we have subsequent task runs check if a classifier already exists (with sufficient training-image count) before proceeding.
- If the train job is collected successfully, the classifier fields get updated with the train results.
- If the train job is not collected successfully due to an error, the classifier does not get updated. It remains in the database with valid = False. So the untrained classifier still blocks subsequent train jobs, unless more images are annotated.

Sources can be manually force-retrained if this situation happens, but obviously we don't want to be doing that all the time.

[...]

PR #409, opened 2021-08-10
Improve handling of training failures + add more Classifier statuses

StephenChan:
Addresses issue #408.

When a training job is collected, if the job failed, delete the corresponding `Classifier` from the database. It seems that we don't have to keep the classifier around when this happens, and deleting it allows a new classifier to be trained automatically (specifically, on the next run of the periodic `submit_all_classifiers` task).

Note: This PR also changes requirements to get pyspacer from GitHub instead of PyPI, so that it's easy to get the latest pyspacer code without having to release a new version. (Incidentally, the AssertionError referred to in issue #408 is only on GitHub at the moment, not PyPI.)


beijbom (approved):

beijbom:
I think it's a good step in the right direction. Obviously, the 1 hour delay is a hack that should be removed. For example, AWS Batch has a system with dependencies between jobs. Ideally we'd embrace that to eliminate these race conditions.

But again, this seems like a good workaround.

project/vision_backend/tasks.py
```diff
@@ -285,14 +285,18 @@ def collect_all_jobs():
 def _handle_job_result(job_res: JobReturnMsg):
     """Handles the job results found in queue. """
 
+    task_name = job_res.original_job.task_name
+
     if not job_res.ok:
         logger.error("Job failed: {}".format(job_res.error_message))
         mail_admins("Spacer job failed", repr(job_res))
```
beijbom, line 292: I wonder if we should skip the `mail_admin` step for the errors we handle? Just to reduce noise?

StephenChan: Probably, yeah - at least for the time being.

If we can later resolve the job order (or something else) so that this AssertionError doesn't happen nearly as often, then it'd be good to turn these admin mails back on, so we can be notified about other types of train/deploy errors.

StephenChan:
Two thoughts:

1) In the last few days, we got two cases on the Google Group where the force-retrain would also fail, unless the images' features were force-reextracted first. In both cases, the source owner happened to add/remove labels from the labelset at some point. I'm not sure what's the chain of events here which leads to the features being left out of date. In any case, I guess we do need to make sure the features get force-reextracted if necessary.

2) Instead of using the 'delete classifier if training fails' solution (implemented in this PR so far), I'm thinking of the following:

   - Add more possible classifier statuses. Currently we have a boolean for valid vs. invalid. Make a DB migration to change the status boolean to a string, with four possible states: training pending, training failed, rejected due to not meeting improvement threshold, accepted.

   - If training fails, instead of deleting the classifier, we mark the classifier status as 'training failed'. Then we check whether the previous N classifiers (N = 3?) for the source all have 'training failed'. If yes, mail the admins with the latest training error. If no, rerun training with a reasonably short delay, say 3 hours.

beijbom:
Yeah. That sounds good. One option is to add that if there are 2 (or even 1?) failed trainings in a row, we force re-extract all features for that source. That will make it more like the next training succeeds.

project/vision_backend/tasks.py
```diff
@@ -114,8 +114,8 @@ def submit_classifier(source_id, nbr_images=1e5, force=False):
     # Create val-labels
     val_labels = th.make_dataset([image for image in images if image.valset])
 
-    # This will not include the one we just created, b/c it is not valid.
-    prev_classifiers = Classifier.objects.filter(source=source, valid=True)
+    # This will not include the one we just created, b/c status isn't accepted.
+    prev_classifiers = source.get_accepted_robots()
```
StephenChan, line 118: @beijbom From your memory, is `prev_classifiers` meant to have the classifiers in a particular order? I changed it to list classifiers from newest to oldest. The old order was arbitrary, as far as I could tell, but I just wanted to double check.

StephenChan:
Finally got around to this PR again.

> Add more possible classifier statuses. Currently we have a boolean for valid vs. invalid. Make a DB migration to change the status boolean to a string, with four possible states: training pending, training failed, rejected due to not meeting improvement threshold, accepted.

Done in commits 4f50b4d, bc9d0e5, 61a0189.

> If training fails, instead of deleting the classifier, we mark the classifier status as 'training failed'. Then we check whether the previous N classifiers (N = 3?) for the source all have 'training failed'. If yes, mail the admins with the latest training error. If no, rerun training with a reasonably short delay, say 3 hours.

> One option is to add that if there are 2 (or even 1?) failed trainings in a row, we force re-extract all features for that source. That will make it more like the next training succeeds.

I put in N = 4, with force re-extracts after 2 failures and after 3 failures. See what `train_fail()` looks like now for details: https://github.com/beijbom/coralnet/pull/409/files#diff-c46386d73f2f26647da3cfa3aae3a8785185a8223d3177708d5199c2f8de9273R364

Also, I rewrote `Source.need_new_robot()`. The only logical change is that it now retries training when the last robot's status is failed. Other than that, I just restructured the method: https://github.com/beijbom/coralnet/pull/409/files#diff-140724231c27f65f47fb2844c0fe3f55838bdb0182331df9aafc5e8405f8d3adR406

Commits e4f3c3e and ec80291 were done to make unit tests easier to reason about. They shouldn't affect non-test behavior; I re-ran one of the regtests as a double check.

@beijbom There was one small thing I wanted to double check with you in `vision_backend/tasks.py`; I left a comment there. Other than that, I think this should be ready to merge.

project/vision_backend/tasks.py
beijbom: You are right in that it doesn't matter. We just compare the current accuracy to max across all the previous ones.

project/vision_backend/tasks.py
```diff
@@ -323,6 +330,83 @@ def _handle_job_result(job_res: JobReturnMsg):
...
+        # Only 1 fail so far.
+        # Sometimes the features are getting re-extracted, and just didn't
+        # finish yet. We'll hope for that here.
+        submit_classifier.apply_async(
+            args=[source.pk],
+            eta=now() + timedelta(hours=3))
+    elif 2 <= train_fails_in_a_row <= 3:
+        # Go through all the source's images and force-extract features.
+        for image in source.image_set.all():
+            reset_features.apply_async(
```
beijbom, line 374: You could also call `submit_features` directly with the `Force` flag set. https://github.com/beijbom/coralnet/blob/master/project/vision_backend/tasks.py#L35

beijbom: Nevermind -- I like your way better.

beijbom:
Thanks @StephenChan . Looks solid. Hope this will do the trick!

StephenChan:
This is now live in production, so we'll see how it goes!

PR #418, opened 2021-10-29
Adding Qimins paper

beijbom:


beijbom:
CC @StephenChan @kriegman .

StephenChan:
OK, so you added the new paper, reordered the other papers, and moved Qimin from 'Current' to 'Alumni' - looks good I think.
How about we add Haoming (Tony) to the alumni list while we're at it?

StephenChan (approved):

beijbom:
Yeah, I ordered the papers by reverse chronological order. Also added Haoming just now.

Issue #412, opened 2021-09-27
Sources with only 1 label keep getting training errors


StephenChan:
Currently happening with source 1833.

`need_new_robot()` returns `True` for the source, since it has the minimum number of annotated images and it hasn't successfully trained a robot yet. However, spacer returns an error every time it tries to train, because the training data only has 1 unique label in it:

```
  File "/workspace/spacer/spacer/train_utils.py", line 58, in train
    raise ValueError('Not enough classes to do training (only 1)')
ValueError: Not enough classes to do training (only 1)
```

It turns out the source's labelset also only has 1 label, but in general, the annotations' contents are the determining factor here.

A few possible ideas:

1. Return from the `submit_classifier()` task if the training data only has 1 label in it. Pretty straightforward. However, since `need_new_robot()` will continue to return True for this source, we'll keep seeing the source highlighted in red in the Backend Overview page.

2. Change `need_new_robot()` to check if the training data only has 1 label in it, as the last check before returning True. I don't think the query is super simple/fast, since we have to check only the annotations in the confirmed images - not just all the confirmed annotations. So it could take on the order of 20+ seconds per small/medium source that needs a robot. This delay would happen each time we open Backend Overview, and during the daily `submit_all_classifiers()` loop (as opposed to during the `submit_classifier()` task).

3. Change spacer's training to handle the trivial case of 1 unique label, returning a valid classifier instead of giving an error.

4. Add another classifier status option, "Not enough labels". Update the `submit_classifier()` task to catch the case where the data only has 1 unique label. When this case is caught, set the classifier status to "Not enough labels", and return from the task without calling spacer. When the latest classifier's status is "Not enough labels", treat it like the "Rejected for low accuracy" case: submit a new classifier when we have 1.1x the number of confirmed images.

Option 3 is probably the simplest option from CoralNet's perspective, so I'd want to lean towards that, but I'm not sure how simple (or possible?) it is in spacer. If Option 3 is not particularly easy, then Option 4 might be the next best.

StephenChan:
Updated option 2's ramifications:

> Change `need_new_robot()` to check if the training data only has 1 label in it, as the last check before returning True. I don't think the query is super simple/fast, since we have to check only the annotations in the confirmed images - not just all the confirmed annotations. So it could take on the order of 20+ seconds per small/medium source that needs a robot. This delay would happen each time we open Backend Overview, and during the daily `submit_all_classifiers()` loop (as opposed to during the `submit_classifier()` task).

[...]

PR #420, opened 2021-11-04
Avoid training errors when there's only 1 unique label

StephenChan:
For issue #412. Went with option 4 as described in that issue.


StephenChan:
One old unit test failed in the PR merge, despite passing before merge. That's annoying. I guess it's an intermittent test failure, I'll probably look into it next week.

Issue #458, opened 2022-10-19
BatchQueue job-collection breaks when batch_token is None


StephenChan:
(tl;dr: Rare bug; very disruptive and requires manual recovery; minimal fix should be easy.)

For context, a `BatchJob` model instance is created when there's a new Spacer job to be run through AWS Batch: feature extraction, classifier training, or image classification.

On September 28, from 9:47 AM to 1:07 PM, 36 of the `BatchJob` instances which were created during this period (21 classification, 15 extraction) were left with a `batch_token` of None.

This was problematic for the periodic job-collection task:

```python
def collect_job(self) -> Optional[JobReturnMsg]:
    batch_client = boto3.client(
        ...
    )
    ...

    for job in BatchJob.objects.exclude(status='SUCCEEDED').\
            exclude(status='FAILED'):
        resp = batch_client.describe_jobs(jobs=[job.batch_token])
        ...
```
    
Since when `job.batch_token` was None, the `describe_jobs()` call would crash with an error like this, as seen in the Celery logs:

```
Traceback (most recent call last):
...
  File ".../coralnet/project/vision_backend/queues.py", line 108, in collect_job
    resp = batch_client.describe_jobs(jobs=[job.batch_token])
...
  File ".../site-packages/botocore/validate.py", line 297, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter jobs[0], value: None, type: <class 'NoneType'>, valid types: <class 'str'>
```
    
And upon crashing, the job-collection task doesn't get the chance to try collecting the rest of the jobs in the queue, even the jobs with valid batch_tokens.

As a result, most classification, extract, and train jobs were blocked up indefinitely. Every few hours, a few jobs would trickle through, since the order of the jobs returned by `BatchJobs.objects...` isn't guaranteed to be always the same. But overall, things were very stuck until I manually marked these 36 jobs as 'failed'.

Google Group thread on the stuck queue: https://groups.google.com/g/coralnet-users/c/7BSazee3jrY

## How batch_token became None

Here's the code that creates the `BatchJob` and later sets the `batch_token`, in `vision_backend/queues.py`:

```python
class BatchQueue(BaseQueue):

    ...

    def submit_job(self, job_msg: JobMsg):

        batch_client = boto3.client(
            ...
        )
        
        ...

        batch_job = BatchJob(job_token=self.get_job_name(job_msg))
[...]

Issue #413, opened 2021-09-27
Some sources seem to get stuck at 1 unclassified image


StephenChan:
This isn't giving us spacer errors, but it's leaving red rows in Backend Overview. See screenshot below - the rows where "# ver rbts" is at least 1, and "# uncl." is also at least 1. I can't think of how this might've happened.

![image](https://user-images.githubusercontent.com/857711/135000730-0deffd5a-4cda-4053-8c05-d700eaaa50d8.png)

StephenChan:
My remark in PR #519 was misleading. It's cool to have "Unclassified, without features" at 0, but what this issue actually wants is "Unclassified, with features" at 0. That would involve separating out sources that don't have a classifier yet.

![image](https://github.com/coralnet/coralnet/assets/857711/6abe137f-ba60-4432-b870-06f337625bc3)

StephenChan:
Well, I haven't noticed this phenomenon recently, and in theory the various changes to detecting source state (such as PR #459 and PR #549) should have solved it.

Issue #159, opened 2018-06-03
Robot status or event log displayed on Backend page


StephenChan:
Possible status messages include 'pending training', 'new robot trained, improved by 5%', 'reclassifying 50 images'.

At minimum this could be a single status message. Or we can have an event log with dates for each event.

StephenChan:
Just want to make sure this is mentioned somewhere: We're often asked why no new robot has been trained for a particular source in a while, even after annotating 1.1x more images. Usually, the reason is that a robot WAS trained, but wasn't saved due to a lack of accuracy improvement over the previous robot. We should start communicating this type of event to the source owners somehow. One way could be using this event log.

beijbom:
Yeah. That was the whole idea behind setting that up. I just never ended up
adding any event messages. :(

StephenChan:
We'll just make this a higher-priority Beta 3 task, I suppose.

Actually, there's something else I realized regarding clarity of the backend info. If you check out this source: https://coralnet.ucsd.edu/source/1656/

- The only valid classifier has 1653 images. There are currently 8049 images in the source.
- The valid classifier is shown as having 70% accuracy, but when we try to train a new classifier, it says `pc_accs` for the valid classifier is 85%. So the new images (or maybe changed annotations) really boosted the accuracy of the valid classifier.
- The confusion matrix shows n = 10144. That's based on 1653 images, since 1653 images / 8 * 50 points per image = around 10k.

So, perhaps the valid classifier is indeed still the best one, but the stats shown (overall accuracy and CM) are the stats from the dataset at the time of original training. Is this by design, or is it something that should be changed? (Or did any of this change in beta2-rollout?)

beijbom:
hey @StephenChan . You bring up a good point. I suppose I wasn't expecting several robots in a row to fail. And perhaps users will find it weird that their robot's performance suddenly changes without a new one being added? But in principal I agree that what you describe is the right way to go. We'd probably have to add one of those event messages at the same time to communicate what happened.

StephenChan:
I think this issue is pretty much covered by the jobs dashboard developed in the last couple of years: https://coralnet.ucsd.edu/blog/jobs-dashboard/

The top status message of the jobs dashboard does include things like:

- Waiting for training to finish
- Scheduled 15 image classification(s)
- Need 120 annotated images for next training, and currently have 110

And also, the dashboard's jobs list logs training attempts from the past 30 days, and next to each attempt there's a description of the result, which may be along the lines of:

[...]

PR #459, opened 2022-12-01
Async task management improvements

StephenChan:
I started off wanting to handle issue #458 (None batch_token when collecting AWS Batch jobs), but was unsatisfied with the code I'd have to write to retry whatever the dropped job was.

I ended up working on various things related to async jobs - status tracking, error reporting, more intelligent ordering of tasks, etc. Some of these ideas had been going through my mind for a while now.

## Improved and encapsulated ordering/timing of async jobs

As an example, we had this part of the upload-annotations view (in `upload/views.py`):

```python
        for image_id, annotations_for_image in uploaded_annotations.items():

            ...
            Point.objects.filter(image=img).delete()
            ...
            Point.objects.bulk_create(new_points)
            ...
            # Submit job with 1 hour delay to allow the view and thus DB
            # transaction to conclude before jobs are submitted.
            # Details: https://github.com/beijbom/coralnet-system/issues/31.
            backend_tasks.reset_features.apply_async(
                args=[img.id], eta=now() + timedelta(hours=1))
```

That final call is now changed to:

```python
            # The uploaded images should be ready for feature extraction.
            queue_source_check(source_id)
```

`queue_source_check()` queues up a `check_source` Job. Before explaining the details of job queueing, I'll show the rough structure of check_source() first:

```python
@task(name="Check Source")
@job_runner()
def check_source(source_id):
    ...
    [If there are images without features:]
        [Queue feature extraction jobs]
        return
        
    [If need new robot:]
        [Queue training job]
        return
        
    [If there are images to classify:]
        [Queue classification jobs]
```

So check_source() looks at the source status as a whole and queues whatever needs to be done next. It only queues one of extraction / training / classification, not a combo, so that things are less likely to run in the wrong order.

Even if the upload-annotations view has a guess that feature extraction is probably the next thing to run, it defers to check_source() to decide what to actually run next.

Also, check_source() will now be run daily for each source. We used to have a daily task which only checked which sources needed new classifiers. This meant that if extraction or classification ran in the wrong order, then images could be left unextracted / unclassified indefinitely. That shouldn't happen anymore. The only thing is that we should keep an eye on the check_source() runtime to make sure it's reasonable.

Coming back to queue_source_check(), it calls `queue_job()`. Here's what's roughly involved:

```python
def queue_source_check(source_id, delay=None):
    queue_job('check_source', source_id, ...)
    

def queue_job(name, ..., initial_status=Job.PENDING):
    ...
    jobs = Job.objects.filter(job_name=name, ...)
    try:
        job = jobs.get(status__in=[Job.PENDING, Job.IN_PROGRESS])
    except Job.DoesNotExist:
        job = Job(...)
        job.save()
    else:
        logger.debug(f"Job [{job}] is already pending or in progress.")
        return None
    ...
```

Putting the existence check aside for now, the primary thing happening here is creating an instance of the `Job` model (defined in a new `jobs` app). A Job tracks the status of any async work unit we care to track. That includes a source_check, a feature extraction, a classifier training, or an image classification (API or otherwise). The idea is to handle these async units in a uniform fashion, so they can all benefit from the same developments in robustness, error handling, etc.

Some Jobs are immediately set as IN_PROGRESS, and others as just PENDING. Here's how PENDING Jobs are run - the `run_scheduled_jobs()` periodic task:

```python
def get_scheduled_jobs():
    jobs = Job.objects.filter(status=Job.PENDING)
    jobs = jobs.filter(scheduled_start_date__lt=timezone.now())
    return jobs


@periodic_task(
    run_every=timedelta(minutes=3),
    ignore_result=True,
)
@full_job()
def run_scheduled_jobs():
    for job in get_scheduled_jobs():
        starter_task = job_starter_tasks[job.job_name]
        starter_task.delay(...)
```

So this task grabs any PENDING Jobs which are due to run, matches the Job with a particular celery task, and runs that celery task.

run_scheduled_jobs() runs in its own thread, separate from any thread that might create a Job instance. Therefore, it can't start a Job's celery task until the Job instance is committed to the database.

So, thinking back to that upload-annotations view, the check_source Job will be unknown to run_scheduled_jobs() until the upload-annotations view completes and the view's transaction is committed. This is why the explicit 1-hour delay is no longer needed. In general, views shouldn't have to worry much about task timing anymore: any Job queued up during the view can't be run until the view's finished.

For reference, the 1-hour delay is from PR #344, and issue #408 is related as well. Those issues show that we put a lot of thought into the upload-annotations timing issue in particular, and hopefully the new Job scheme makes this type of concern simpler to reason about.

## Robustness against race conditions

The reliance on check_source() already addresses at least one race condition concern - ordering extraction vs. training. There are a few more things about the Job design which minimize race conditions:

- queue_job() has guards against creating/running duplicate Jobs, e.g. two training Jobs for the same source, or two extract Jobs for the same image.

  I cut out the details above, but the guard against duplicate in-progress Jobs is a database-level uniqueness check, making it as robust as possible against race conditions. This guard doesn't apply to Jobs that are already completed, so the last completed training Job for source X doesn't have to be cleaned up before running source X's training again.
  
- After a Job is queued as PENDING, it needs to be started later. This is handled by `start_pending_job()`:

  ```python
  def start_pending_job(job_name, arg_identifier):
      # select_for_update() locks these DB rows from evaluation time
      # to the end of the transaction. This ensures that we don't
      # have two threads starting the same job at the same time.
      # Instead, the first thread will get to run, while the second thread
      # will get a DatabaseError and return.
      # (This works most effectively if we're using autocommit when entering
      # this function.)
      jobs_queryset = Job.objects.select_for_update(nowait=True).filter(
          job_name=job_name,
          arg_identifier=arg_identifier,
          status__in=[Job.PENDING, Job.IN_PROGRESS],
      )
      with transaction.atomic():
          # Evaluate the query.
          try:
              jobs = list(jobs_queryset)
          except DatabaseError:
              # Probably tripped the select_for_update() row locking, meaning
              # there's another thread in here already.
              logger.info(
                  f"Job [{job_name} / {arg_identifier}] is locked"
                  f" to prevent overlapping runs.")
              return None

          ...
          job.status = Job.IN_PROGRESS
          job.save()
      ...
  ```
  
  `QuerySet.select_for_update()` is probably the main way of doing database row-locking in Django. The idea here is to prevent two separate threads from simultaneously changing a single Job from PENDING to IN_PROGRESS.
  
  So we have this guard against multi-starting Jobs, and we have the earlier guards against Job duplication. There's one more piece:
  
- Decorators like `@full_job` and `@job_runner` can be placed over functions to track them as Jobs. The reason run_scheduled_jobs() is itself tracked as a Job is that it benefits from the guarantees against duplicate runs. This way, there's no danger of a single Job's celery task being started twice simultaneously (by different runs of run_scheduled_jobs()). Similarly, `collect_spacer_jobs()` is itself a Job, so that a single spacer-job can't be collected twice.

  These three pieces together give us a good guarantee that the same Job isn't going to have simultaneous runs, which should cut out a host of potential race conditions.
  
- A limitation: `reset_features()` and `reset_classifiers_for_source()` aren't currently tracked as Jobs, as I wasn't sure how to mesh those tasks well with check_source(). So this leaves a potential race condition along the lines of:

  1. Submit feature extraction
  2. Generate points + clear features (e.g. point regeneration)
  3. Spacer runs feature extraction on old points
  4. Feature extraction on old points is collected
    
  There's a check in the feature collection code (step 4) which checks if the points are out of sync. So the remaining concern comes when steps 2 and 4 overlap in separate threads. More specifically:
        
  - Thread 4: rowcols = old Points from DB; they match the feature rowcols, so proceed
  - Thread 2: commit new Points; img.features.extracted = False
  - Thread 4: img.features.extracted = True
  
  And so the image ends up marked as having features, but they don't match the new Points. However, the race condition to get this to happen appears to be quite fine-grained, so I'm willing to wait and see if it even comes up.
  
## Consistent error handling across a Job's lifetime

We have Jobs that run in three phases: a pyspacer-submission celery task, in AWS Batch running pyspacer, and then a pyspacer-collection celery task.

I wanted errors of all three phases to report back to the Job's status and result-message fields, so I tried to unify the error handling between these different parts of the code. One common theme is that a `JobError` can be raised at various parts of the code, to indicate a (generally anticipated) situation which should result in the Job being marked with FAILURE status.

There's also support for catching unanticipated errors. This results in FAILURE status, creating an ErrorLog instance, and mailing the admins - much like a 500 Internal Server Error in our synchronous views. Previously, these uncaught errors in tasks could easily slip through the cracks, since they were only detectable by checking celeryworker log files.

In this part of a Job-decorator's code, `report_task_error()` is what handles the unanticipated-error situation:

```python
            success = False
            result_message = None
            try:
                result_message = task_func(*task_args)
                success = True
            except JobError as e:
                result_message = str(e)
            except Exception as e:
                # Non-JobError, likely needs fixing:
                # report it like a server error
                result_message = str(e)
                self.report_task_error(task_func)
            finally:
                # Regardless of error or not, mark job as done
                finish_job(job, success=success, result_message=result_message)
```

Jobs also have an `attempt_number` field which is incremented each time the same job fails in a row. When it gets high enough (currently 5), admins are mailed. We had this sort of logic for classifier training, but now it applies to all Jobs.

## BatchQueue changes

As part of improving error handling, and the issue #458 error in particular, I did several things with BatchQueue.

- This is the part that handles issue #458, which ended up being a bit more subtle than expected (as explained in the issue comments):

  ```python
          # Iterate over not-yet-collected BatchJobs.
          for job in BatchJob.objects.exclude(
              status__in=['SUCCEEDED', 'FAILED']
          ):

              if job.batch_token is None:
                  # Didn't get a batch token from AWS Batch. May indicate AWS
                  # service problems (see coralnet issue 458) or it may just be
                  # unlucky timing between submit and collect. Check the
                  # create_date to be sure.
                  if timezone.now() - job.create_date > timedelta(minutes=30):
                      # Likely an AWS service problem.
                      self.handle_job_failure(
                          job, "Failed to get AWS Batch token.")
                      job_statuses.append('DROPPED')
                  else:
                      # Let's wait a bit longer.
                      job_statuses.append('NOT SUBMITTED')
                  # In either case, continue to next job.
                  continue
                
              ...
  ```

- The `BatchJob` model now has a foreign key to Job, so that the Job's status can be updated if a problem is found with the BatchJob.
  
- I noticed that `collect_all_jobs()` (now called collect_spacer_jobs()) was checking in-progress Batch jobs more than necessary. Here's how it was:

  ```python
  def collect_all_jobs():
      logger.info('Collecting all jobs in result queue.')
      queue = get_queue_class()()
      while True:
          job_res = queue.collect_job()
          if job_res:
              th.handle_job_result(job_res)
          else:
              break
              
              
  class BatchQueue(BaseQueue):
      ...
      def collect_job(self) -> Optional[JobReturnMsg]:
          ...
          for job in BatchJob.objects.exclude(status='SUCCEEDED').\
                  exclude(status='FAILED'):
              [If AWS Batch says succeeded:]
                  job.status = 'SUCCEEDED'
                  job.save()
                  return return_msg
              [Else if AWS Batch says failed:]
                  job.status = 'FAILED'
                  job.save()
              [Else, running or other status: continue to next iteration]
  ```
  
  So `BatchQueue.collect_job()` was looping over the available Batch jobs until it found a success, at which point it would return. Along the way, it would mark failed jobs as failed, so that those jobs wouldn't get revisited.
  
  However, other job statuses like running would get skipped... and then revisited each time BatchQueue.collect_job() gets called in the `while True` loop, i.e. potentially once for every Batch-job success.
  
  I ended up replacing this with a `BatchQueue.collect_jobs()` generator function. It goes through each Batch job exactly once, and uses the `yield` keyword each time it finds a successful job. And it's called only once by collect_spacer_jobs(), rather than being called in a while loop.
  
- There's now a job status summary which is logged when collect_spacer_jobs() concludes, which looks something like "6 RUNNING, 13 STARTING, 2 SUCCEEDED".
  
- There are now tests for the BatchQueue class, utilizing a mock of what `boto3.client()` would normally return, so that AWS isn't required for these tests.

## Job dashboards

Last but not least, dashboard UIs. This is available from Admin Tools:

Click a source name to list jobs for a source:

check_source jobs seemed to clutter the main table, so instead I put the result message of the latest check_source at the top of the page.

From the main dashboard, click the non-source jobs link to go here:

Deploy jobs are included here since they're not involved in the 'progression' of a particular source.

Paginated with 100 jobs per page. Jobs are purged after 30 days, with the exception of non-error training Jobs (both accepted and rejected), which are kept indefinitely.

The source jobs list is also reachable for source members with Edit or Admin permissions. The source menu at the top has a new blue button "Jobs", and there's also this link below the classifier graph:

There are likely improvements that can be made, but it feels like a decent start, and way better than not having these dashboards at all.

Now people can actually check when training is in progress, rather than saying "I think the threshold for the next training has been met, so it's probably happening now?". And by "people" that includes me, without having to SSH into the server and type a bunch of commands.

## Other stuff

- vision_backend management commands have been refreshed (and most of them now have unit tests).

  Extract and train commands previously had a `force` option which is now redundant. Now it's just: if you don't want to force, then call the check_source command, else call the extract / train commands.

- Each `ApiJobUnit` instance now proxies most of its fields to a Job instance to cut down on redundant definitions (modify date, status, result message, etc.). I should say I'm not too happy with the model naming as is (Jobs being on the same level as ApiJob...Units?), but I couldn't think of a better naming scheme that seemed worth pushing right now.

  I was similarly looking into transferring status tracking from Classifier to Job. However, I couldn't come up with a satisfying way of designing this, so I backed off. One problem is that Classifier has at least one status code which actually says why it failed ('lacking unique labels'), so that would have to be some kind of special case in Job's status.
  
- We should update from Python 3.6 when we can (see the commit which mentions GitHub Actions).

- Overall, issues that are potentially resolved: #458, #413, #159, beijbom/pyspacer#38

- I'll update the production server when I'm up for it, which may not be right away. There are a good number of migrations to run, and there will likely be situations to monitor. Either with things working at all, the tweaked timings of the periodic tasks, error reporting frequency, etc. Also unsure if I want to lump this in with related updates, such as upgrading celery.


StephenChan:
OK, think this is ready. It ended up being a lot.

PR #523, opened 2024-01-10
vb_inspect_extracted_features: Do a different check for legacy features

StephenChan:
Legacy features (i.e. extracted before coralnet 1.0) would always fail the check since the command checked for rowcols unconditionally. This command had this issue since coralnet 1.7, so it's somewhat coincidental that it only came up now.


StephenChan:
Ideally this command would just use the same code as the "sanity check" in pyspacer's `train_utils.py`: https://github.com/coralnet/pyspacer/blob/0.7.0/spacer/train_utils.py#L201

But that requires a refactor on pyspacer's side to put that check in its own function, and I don't feel like doing another pyspacer release right this moment.

PR #534, opened 2024-03-08
Define a view-scoped cache in memory, and have label_details use it

StephenChan:
Turns out that fetching from the disk-based cache can be a bottleneck if we have to fetch the same large structure 16,000 times in a row.

That is what the label-list view currently does. It fetches the label-details structure twice per label (and there are 8000+ labels in production). For example, the code for getting any label's popularity (to show the colored bar on the label list) involves fetches the label-details structure from the disk-based cache. Those fetches apparently add up to incur 10-20 minutes of load time. It's still better that the popularities are cached at all, so we don't have to compute the popularities for every label on every page load, but it's still much slower than desired.

There were multiple ways of fixing this, but I opted to define another layer of caching: a "view-scoped cache" which is in-memory (thus, faster than the disk-based Django cache), created at the start of a web request, and cleared after handling the request. With this setup, the disk-based Django cache saves us from computing the popularities very often, and the view-scoped in-memory cache saves us from having to fetch the label details from disk more than once per view.

This might raise the question of why we don't use a single memory cache (Redis, Memcached) to try to get the best of both worlds. That might end up being a route I take later if there are more caching pains. But for now, I still find that a disk-based cache smooths things out on the testing, development, and maintenance side; for example, cached values that take a while to generate don't have to be re-generated when the server's restarted. I was also thinking that the view-scoped property of this new cache can be quite useful in other situations, such as caching the results of database queries.

### DIFF
--- a/project/config/settings.py
+++ b/project/config/settings.py
@@ -875,6 +875,8 @@ MIDDLEWARE = [
     # each request to understand usage patterns. (ID, not username, to keep
     # things relatively anonymous)
     'lib.middleware.ViewLoggingMiddleware',
+    # Provide a cache which persists for the duration of the view.
+    'lib.middleware.ViewScopedCacheMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
 
     # django-reversion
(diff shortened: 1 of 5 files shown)


StephenChan:
Label list now takes 5 seconds to load instead of 10-20 minutes.

Issue #537, opened 2024-03-25
Let source A's classifier be used by source B (baseline functionality)


StephenChan:
There are several ideas going on in issue #150 - I wanted to make this separate issue to establish some baseline functionality (and associated limitations) as a more achievable issue in the short term. Here's an idea for that:

- Source B can only specify the desired classifier ID at source-creation time. When this happens, it automatically sets up source B's labelset to match source A's labelset.

- There will be a check to ensure that source B's creator has view-access to the source which owns the specified classifier. However, the interface for choosing the classifier might not be intuitive or informative yet. Perhaps just a number-input field for the ID.

- Source B won't be able to specify a different classifier after source creation. Source B won't be able to change its labelset after creation, unless the classifier gets deleted (e.g. source A's labelset changes), in which case source B permanently reverts to a "traditional" source which trains its own classifier.

- This feature would be available to anyone who can create a source (so, any registered CoralNet user), but it would have a note that it's in beta phase and has the limitations laid out above.

Note that Source B would point to a classifier ID, not a source ID, so that Source B doesn't worry about the classifier or the labelset changing from underneath it. (Again though, it's another story if the classifier's deleted - which happens if Source A changes its labelset, changes its feature extractor, or is deleted.)

kriegman:
So, this seems like a version of a published "robot" for source A that is
used by source B. A published robot (say from source A) will basically be a
specific classifier, which implies a specific labelset and feature
extractor. So, in this scenario, source B is using a robot from source A.
This issue, might be addressed in the context fo "published robots" and a
more involved semantic of a hierarchy of sources or source inheritance.

StephenChan:
Right, I edited in this sentence to clarify what types of changes in source A would upend source B's setup:

> (Again though, it's another story if the classifier's deleted - which happens if Source A changes its labelset, changes its feature extractor, or is deleted.)

Yes, maybe it makes the most sense to consider source B as inheriting from source A in this case. Though if we could later make this setup more flexible, saying that source B will be able to change which classifier it points to (even a classifier of a different source), then we might conceptualize it differently.

StephenChan:
Done in PR #549:

> * Source B can only specify the desired classifier ID at source-creation time. When this happens, it automatically sets up source B's labelset to match source A's labelset.

Went beyond this, making the classifier ID specifiable in both the new and edit source forms.

[...]

Issue #306, opened 2020-04-05
Date display behavior should be well-defined and tested


StephenChan:
It'd be nice to settle on answers for the following:

- Do we display dates in the server's local time? UTC time? In the user's timezone?
- What's the best way to detect the user's timezone, if we go that route?
- Should dates be formatted only one way, or should we detect user locales and match those?

And then the behavior should be consistent among the following: source creation dates, annotation history dates, blog post dates, newsfeed item dates, etc.

Unit tests are always good as well.

Issue #415, opened 2021-10-27
UI to replace source's annotations of label A with B


StephenChan:
For example, replace all 'Pavement' annotations in a particular source with 'Noncoral' annotations. The Noncoral label should already be added to the source's labelset in order for this to work.

We already have a Django management command for this - see PR #290. We then said on the Google Group once or twice that we'd take people's requests to run this management command on their sources. However, it'd be better to let source admins run this function themselves somehow.

I think the best place for this function would be the Browse Patches page, since this allows for further filtering of the annotations if desired; for example, one could run this only on Unconfirmed annotations, or only on annotations done in June 2021 by a particular user. So you'd apply the filters, then there'd be an action box at the bottom of the page which lets you perform an action on the selected annotations - for now, the only available action would be changing the selected annotations to a particular label. Since this could be a very large batch action which could be very difficult to reverse, there should be an 'are you sure' confirmation dialog.

If you remember issue #44, which we've thought of doing many times over the years, there is definitely some overlap there. Essentially, right now I'm proposing adding the 'assign label' action, without bothering with any of the other ideas for the Browse Patches page yet:

![slide-8](https://user-images.githubusercontent.com/857711/76151742-1c41d680-606d-11ea-8322-c316b1d6d03a.png)

StephenChan:
Another use case here: https://groups.google.com/g/coralnet-users/c/2qChzQzTd1k

> I have a number of unknown or unconfirmed species in my samples - and I usually Label them Unknown 1/2/3 etc until I confirm the identity. So this now poses a question .... can you change the name of a label once you have made a positive ID? and if you change the label does it then change the label globally in your source so that images that used to be labelled Uknown 1 are now given their correct name?
>
> Of course, if you made an error with your identification this would also apply if you were changing one species label for another ...

PR #549, opened 2024-05-13
Let one source use another source's classifier

StephenChan:
For issue #537. So far this PR just implements the source state logic, without adding/revising classifier usage semantics yet (that is why this PR's still WIP).

The work so far includes changes to new/edit source forms, handling cases like classifier deletion and labelset editing, and unit tests thereof.

Also revised the source main page layout, mainly to try and group all the classifier related info together; might have more ideas for improvement later, but this should do for now.

Screenshots follow; first, the new/edit source form's option for the classifier mode:

source-form_train

source-form_use-existing

One possible error when submitting this form; there are also errors for not having permission to the classifier's source, or for the classifier not existing:

error_labelset-mismatch

Updated layout to the source main page:

source-main

Three possible classifier status displays; don't mind the "minimum of 3", that's just the setting I use for testing:

status_train

status_train-none

status_use-existing

And two more. The first is when you had a classifier selected, but then it got deleted (say, because the classifier's source changed its labelset or feature extractor). The second is how it looks when you leave the classifier ID field blank; this would essentially turn off automatic classification, in case that's ever desired (and for technical reasons, I think it's also a useful default state to have).

status_deleted

status_none

And when you try to edit your labelset when using another source's classifier, it just doesn't let you. Maybe this can be made more lenient later (like allowing the labelset to be a superset, instead of saying it must be identical), but opting to keep it strict, safe and simple for a start.

labelset-edit-denied

Meanwhile, made a couple steps towards issue #306 (making date displays consistent); lib.utils date_display() and datetime_display() both go for the server's local time. But in the future we may want to detect the user's timezone or let them set it.


Jordan-Pierce:
@StephenChan 

Just lurking, but thoughts for the following:

"And when you try to edit your labelset when using another source's classifier, it just doesn't let you. Maybe this can be made more lenient later (like allowing the labelset to be a superset, instead of saying it must be identical), but opting to keep it strict, safe and simple for a start."

What about having an editable mapping between the labelset of the Classifier's Source and the users' Source labelset? So if my Source has "Acropora" in it, but the classifier I want to use has "Acropora Sp." (or something equally as trivial), instead of adopting the classifier's labelset, before submitting a request to classify I can choose how I would want to map each the classifier's output labels to the labels in my Source's labelset? This would be useful for these situations (where the naming convention is just slightly different), but also for more extreme cases as well, depending on the user's needs.

Just a thought. Looking forward to the new feature!

StephenChan:
Hmm yeah, maybe. We try to discourage duplicate labels on CoralNet, but the reality is that some of them have stuck around in prominent sources. (Doesn't help that we never got around to issue #415, but still.) Also, an editable mapping could have a few possible uses, including letting you map two or more labels of the classifier's source to a single label of your source.

Jordan-Pierce:
> Hmm yeah, maybe. We try to discourage duplicate labels on CoralNet, but the reality is that some of them have stuck around in prominent sources. (Doesn't help that we never got around to issue #415, but still.) Also, an editable mapping could have a few possible uses, including letting you map two or more labels of the classifier's source to a single label of your source.

Agreed, the latter would also allow a user to use multiple classifiers (if the functionality is provided), and then map their each of their labels to a common label that they desire.

StephenChan:
- Updated the backend logic (feature extraction, training, classification) to support the new classifier options.

- Changed the `deployed_classifier` field semantics so that it's always the classifier being used for classifying that source's images. So, now it's auto-populated whenever the `trains_own_classifiers` mode is True. I thought this scheme seemed a little simpler compared to only looking at `deployed_classifier` when the train mode is False. With this change, we can check `deployed_classifier` to see what is being used for classification, without caring about whether we're in train mode or not.

- Revised the wording of help texts to better suggest that you can use the `deployed_classifier` field to choose a previously trained classifier in your own source. This capability didn't require any special code. And I didn't see a reason to delete the trained classifiers when toggling train mode off or on, so that's not an obstacle.

- Updated the descriptive text of the currently deployed classifier, including saying the train date, and saying whether it's the latest classifier of its source (and if not, saying what the latest classifier is).

source-edit

source-main

In this PR overall, I kept making revisions and adding tests here and there as I thought of them. So it's likely that the code can be better organized and that I overlooked some situations. But I don't know if I can improve it by staring at the code more right now, so I'll just keep an eye on things after updating the server. I'll also think about how the displayed info in various cases can be improved.

That's it for this PR; will have to think about the label mapping idea more, to see if it's the best way to meet the relevant needs, and to consider what the UI might look like.

StephenChan:
> Does this new feature currently allow for users to train on top of the existing classifier that they choose? Or does the source at this point serve as a place for users to upload images (and point locations), for which an existing classifier will label?

Just the second one.

> What happens after a user adds additional labels to the source that would normally trigger a (re)training?

Nothing in particular. However, if you later decided to switch back to the 'Train' classifier mode, then a (re)training would happen.

And good to hear!

PR #553, opened 2024-06-24
ErrorLog and backend-main-form fixes

StephenChan:
Fixes a few potential server errors that have come from hacker-bot requests this weekend. These errors don't actually indicate vectors for hacking, but getting thousands of error emails due to these bots presents a maintenance problem and distracts from other errors.


StephenChan:
Seems this did the trick, since the server errors were still happening extremely rapidly until I updated the server with this PR, at which point they stopped, according to errorlogs on the admin UI.

PR #559, opened 2024-09-26
Prep for version 2 of calcification rate tables

StephenChan:
- Updated the logic of global rate tables, so that we can provide both v2 and v1 tables in the export options, while only showing the v2 tables' rates on label-detail pages.

- Improved column widths on the 'manage label-rate tables' dialog. Previously, the Name column wasn't wide enough, making the table taller than it needed to be with the global-table names we use.

- Updated help text. Here's the help text from the label detail page (the table names with "(sample)" are placeholders) (EDIT: Screenshot updated 2024/12/09 to reflect updated text):
  
  label_detail__rev2

  The help text for the calcification rate export form has the following section at the start, and after that it has the same help text as the label detail page:

  export

The actual v2 rate tables aren't uploaded to the site yet. Once this pull request's changes are reviewed and merged, and those changes are pushed to the CoralNet website, then I'll upload the v2 rate tables.


StephenChan:
And here's what the label-rate table manager is planned to look like once the v2 tables are uploaded - utilizing the description field to make it clearer that the v2 tables are newer:

table

StephenChan:
Unfortunately dropped this for a couple months. But now after git-rebasing, and applying wording edits from email discussions, this should be good to go for the next production update.

PR #562, opened 2024-10-15
No more daily source checks, sitewide job-status page, and more

StephenChan:
- No more daily source checks. I was intending to profile which job types use the most runtime, but from just eyeballing the job statuses in recent weeks, it seems like 90+% of the time it's the source checks that are running. So I think it's safe to say that we've grown out of the point where "automatically check every source daily" is reasonable. For the few cases where we actually want to check every source, there's now a management command for that, `vb_check_all_sources`.
- Provide a button for source admins/editors to run a source check on-demand. There's a cooldown of 30 minutes. Hopefully this isn't needed too often, since source checks are still automatically scheduled upon certain events; such as images being uploaded, or the last of some feature-extract jobs getting finished. But when it is needed, the capability's there. This on-demand button, on average, should be a far less resource-intensive fallback protocol compared to "check every source daily".
- Add a sitewide job-status page, where it shows recent average wait times and the number of jobs in the queue over time. Viewable for anyone who can log in.
- Tweaks/enhancements to the job list views.
- New help dialog for the source job list.


StephenChan:
Seems that we're finally getting the incomplete jobs under control.

image

image

PR #565, opened 2024-10-16
Job monitoring tweaks + Event model migration and indexes

StephenChan:
Jobs:

- Job tables: include runtime in timing info for completed jobs.
- Job status page: show longest pending or in-progress time to admins, instead of just showing longest pending time. Since jobs can get stuck at either of those statuses.
- Backend overview: sources that haven't been checked in the last 30 days are considered normal and not a concern now, so don't color them blue (just apply no color) and move them to the bottom of the table.

Events:

- Add an Event subclass for source check requests. These events aren't displayed anywhere yet, but can be helpful later to get a sense of how often folks have to request source checks themselves (and in what situations).
- Move ClassifyImageEvent from events to vision_backend. As we go on to add more types of events, it seems more manageable for each subclass to live in the app that relates to the type of event it is.
- Add DB indexes for Event. During a source check, the ClassifyImageEvent lookup alone could apparently take 2-3 minutes. Hopefully this speeds things up. The table lacked indexes because the references to source and other tables aren't actually foreign keys.


StephenChan:
> Add DB indexes for Event. During a source check, the ClassifyImageEvent lookup alone could apparently take 2-3 minutes. Hopefully this speeds things up.

I'm not sure if my characterization of source check runtimes was accurate, or skewed by focusing on very large sources. I'm also not sure if source checks sped up at all; oh well.

PR #566, opened 2024-10-21
Improve classification scheduling and performance

StephenChan:
Although the job queue's been relatively stable in recent days, I noticed times where a single source was using all the background-job resources for 5 or so hours uninterrupted. It was something like 1000 images with 400-1000 points each, all getting queued up to be machine-classified.

A single classification with an average amount of points is quite fast (1-2 seconds), which is why we don't farm classifications out to AWS Batch. However, at 400 points, these were taking like 15 seconds. There's also the matter that CoralNet lets a source queue up as many classifications as it can queue up in 10 minutes (the enforced timeout for the check-source job), which is why it can be thousands of images in a row from the same source.

High point count or not, classifications are - as far as I can think of - currently the biggest way that a single source can monopolize the background-job processes. Which means no harm to webpage responsiveness, but delays any sort of vision-backend work for every other source. So, this PR's main aim was to limit this monopolization potential, and also addresses related concerns.

Details:

- Now when a source check schedules classifications, it stops after it schedules a certain amount of classification "work". The "work score" is computed as 100 per image, plus point count for each image. If a single source check reaches 100,000 work score, then it stops scheduling classifications, and you generally have to wait till the previous classifications are finished before another source check will run.

  So at 10 points per image, you'll be limited to around 900 images in a row, at 100 ppi you'll be limited to 500 images, and at 1000 ppi you'll be limited to 90 images. Also, once the last classification is finished, the following automatic source check has a delay of 10 minutes, which should allow other CoralNet background jobs to get a chance.

- When testing, I realized that this part of the classification scheduling logic was no longer particularly accurate:

  > Here we detect whether the deployed classifier is the latest one used for ANY classifications. We check this because, most of the time, a given classifier has either taken a pass on ALL classifiable images, or it hasn't run at all yet.

  In fact, it didn't just become inaccurate now. It also wasn't accurate when I introduced the 10-minute timeout to source checks in PR #480 (part of production starting November 2023). If the check timed out, then the deployed classifier would be the latest one among any classifications, but it wouldn't be queued to run on all classifiable images.

  So I ended up introducing a new DB field, `ImageAnnotationInfo.classifier`, to track the latest Classifier that took a pass over each Image. Assuming this field gets updated when it should, this makes the images-to-classify logic more accurate as well as a lot simpler.

  The new field is pre-populated by a data migration, but only for images which have ClassifyImageEvents associated with them. These events were introduced in CoralNet 1.7. So, any images that were last classified before 1.7 are subject to a redundant re-classification on the next source check. However, since 1.7 was 11 months ago at this time of writing, it doesn't seem like this will trigger a concerning number of redundant classifications (given that we no longer try to check every source every day).

- I was thinking about what was making 400-point classifications take 15 seconds. At first I was going to farm out these high-point classifications to AWS Batch. But then I remembered that just saving the annotations is slow; they are saved one by one since saving in bulk would skip the django-reversion signals needed for creating Annotation History entries. Indeed, clicking Save in the Annotation Tool also takes at least several seconds with around 200 points.

  But the ClassifyImageEvents introduced in CoralNet 1.7 were already a first step towards reworking Annotation History, such that we'd be able to save in bulk. So I took a few next steps:

  - Instead of just using the ClassifyImageEvents internally (for source checks), the events that are created from now on have their details displayed on the Annotation History page, in the same format as the previous history entries.

  - django-reversion entries are now turned off for machine-classification, so that we don't have duplicate entries along with the Event-derived ones. However, django-reversion is still used for the Annotation Tool and Alleviate.

  - Machine-classification of images now uses `bulk_create()` for a speedup, in the most common case where the image starts off completely unclassified (with zero annotations). Later, we could try to get fancier and bulk-create or bulk-update in other cases, but I thought that most of the time we'd either be creating all the annotations or updating just a few.


StephenChan:
I think classification performance sped up by a factor of less than 2, if anything. Hard to say beyond that. But the scheduling is improved at least.

PR #575, opened 2024-12-08
Handle more cases of pyspacer invalid/desynced rowcols

StephenChan:
Three commits for pretty much three separate cases. These address most of the job errors and error-notification emails that I've been seeing lately.


StephenChan:
Both auto-recovery cases added in this PR got successful examples today.

Issue #111, opened 2017-08-21
Export image covers error: 'NoneType' object has no attribute 'get_locals_ordered_by_group_and_code'


StephenChan:
Example: `AttributeError at /source/<id>/export/image_covers/
'NoneType' object has no attribute 'get_locals_ordered_by_group_and_code'`

This server error was seen 4 times this year. I've neglected to investigate it at all, so that needs doing.

StephenChan:
I accidentally got this error when testing export image covers without a labelset. So maybe it's just that.

StephenChan:
Still exists, and also have just seen this with the calcification stats export.

StephenChan:
PR #581 should prevent this from happening by not allowing the export when there's no labelset.

PR #581, opened 2024-12-15
Much faster annotation exports + check for labelset before exporting

StephenChan:
Here are some annotation export test results, run in staging (PPI = points per image):

| Test | Old code | New code |
| --- | --- | --- |
| CSV: 9 PPI, 310 images, no optional cols | 21.5 sec | 3.5 sec |
| CSV: 9 PPI, 310 images, all optional cols | 53 sec | 3.5 sec |
| CPC: 9 PPI, 310 images, no prior cpc uploads | 20.5 sec | 3 sec |
| CSV: 200 PPI, 468 images, all optional cols | 38 min | 14 sec |
| CPC: 200 PPI, 468 images, no prior cpc uploads | 8.5 min | 6.5 sec |
| CSV: 100 PPI, 2072 images, all optional cols | - | 50 sec |
| CSV: 10 PPI, 35413 images, all optional cols | - | 5.6 min |

Not going to lie, feeling pretty good about this! It was all about paying attention to the database queries. Essentially, now it's a few queries per image instead of a few queries per point.

I ran most short tests twice in a row (to check for caching effects), and there was never a significant difference between the first and second tests. The .5 second times are from averaging the two times. I was too lazy to get actual times for the last two tests with the old code, but the last one is the kind of scale that was taking hours in production.

The sped-up exports are Export Annotations, CSV and Export Annotations, CPCe. Export Image Covers and Export Calcification Rates were already on the faster side (I think after this update they are still the faster exports, measuring per-image).

---

Besides performance, each export type now requires a labelset; when you don't have a labelset, the export forms are replaced with placeholder text "This action isn't available because the source has no labelset." The main benefit of this is to prevent the error "'NoneType' object has no attribute 'get_locals_ordered_by_group_and_code'" when export image covers or export calcification rates are attempted without a labelset (issue #111, a long-standing one).


kriegman:
Nicely done Stephen! — David

PR #597, opened 2025-02-20
API: user-status endpoint + speedups + rate limit changes

StephenChan:
Will write details later; seeing if this can be potentially pushed tonight (hopefully my impatience is not punished).


StephenChan:
Added details to first post.

PR #598, opened 2025-02-20
Fix ZeroDivisionError in backend_overview when no photos are in the database

Virtlink:
This PR fixes a `ZeroDivisionError` in the `backend_overview` page when there are no photos in the database.

```
ZeroDivisionError at /backend_overview/

division by zero
```


StephenChan:
Thanks for the suggestion and pointing out the bug. Instead of adding a condition to the calculation though, I think it'd be a bit more simple to short-circuit out of the view function, which we might as well do since the page has no useful info with zero images - see PR #606.

Virtlink:
Okay makes sense, thanks!

PR #604, opened 2025-02-27
collect_spacer_jobs optimizations

StephenChan:
As a result of the API optimizations in PR #597, I noticed a stretch of maybe 10 collect_spacer_jobs runs in production where it was collecting an average of 300 or so jobs per run, and the runs (which are supposed to go every 2 minutes) were falling slightly behind steadily (meaning they were taking an average of 2+ minutes each) until the job demand decreased again. I don't think we've sustained that rate for very long anytime recently, so this situation hasn't caused a noticeable site slowdown. But given that we're trying to raise our API usage capacity, possibly by a lot, it seemed like an imminent bottleneck to address.

Two types of optimizations I thought to make here, needing moderate effort:

- Requesting AWS Batch job statuses is done in batches of 100 at a time (the max that boto allows for one request) instead of one at a time.

- Optimized database queries done in collect_spacer_jobs:

  - Collecting classify (API deploy) jobs now uses less than O(n) queries.
  - Collecting feature-extract jobs still has an O(n) factor, seemingly 3 queries per image, as I've pointed out in the corresponding queries-test class.
  - Collecting train jobs isn't optimized at all because I don't expect there to be many train jobs to collect at once.

Comments:

- Retrieval of job result messages from S3 is still done one image at a time, and I'm not sure how much that might be a bottleneck right now. If needed, one potential road to addressing that is to batch up multiple extract/classify jobs into a single spacer request. Part of spacer's design already has that capability in mind, but since CoralNet hasn't used that at all yet, it's not battle-tested yet.

- Another potential API bottleneck area to keep an eye out for is the task which submits the deploy jobs to AWS Batch. Haven't measured it yet, though.

- Kind of unsatisfying how this led to adding 200 or so lines of code to task_helpers.py and queues.py combined, but oh well. Maybe later there'll be ideas to make it more concise.

### DIFF
--- a/project/annotations/tests/test_delete.py
+++ b/project/annotations/tests/test_delete.py
@@ -5,8 +5,7 @@ from django.urls import reverse
 
 from jobs.tasks import run_scheduled_jobs_until_empty
 from lib.tests.utils import BasePermissionTest, ClientTest, spy_decorator
-from vision_backend.tests.tasks.utils import (
-    BaseTaskTest, do_collect_spacer_jobs)
+from vision_backend.tests.tasks.utils import BaseTaskTest
 from visualization.tests.utils import (
     BrowseActionsFormTest, BROWSE_IMAGES_DEFAULT_SEARCH_PARAMS)
 from ..managers import AnnotationQuerySet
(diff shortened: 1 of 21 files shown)


StephenChan:
After merging this, collecting 300 jobs now takes about 17 seconds, rather than close to 2 minutes.

PR #606, opened 2025-02-28
Backend overview: short-circuit if there are zero images

StephenChan:
Follow-up on PR #598, with a different approach - short-circuiting out of the page instead of getting the stats to work with zero images.

Now in this case, visiting the URL should just show a single line: "There are no images on the site yet, so this page has no useful information to show."


StephenChan:
The test failure seems to be an unrelated (and intermittent) one.

PR #609, opened 2025-02-28
[Temp] Backend overview zero-images fix on top of intermittent test error fixes

StephenChan:
Let's see if this'll ultimately let me do the next merge with passing tests.


StephenChan:
Hopefully that's a yes.

Issue #437, opened 2022-02-02
Use GET HTTP method instead of POST for Browse search


StephenChan:
Been on my mind for a while now, but I don't think there's an issue for this yet. Advantages:

- Page refreshes and Back/Forward button usage are clunky with POST.
- When Back/Forward button usage is less clunky, it would be easier to retain the search params during annotation tool prev/next navigation, and image details prev/next navigation (not implemented yet; issue #438).
- GET means search params will be in the URL, which means it'd be possible to jump to any page number by modifying the page param in the URL.
- It's more correct for this to be GET rather than POST, since no data is being modified. We'll particularly want to use the correct HTTP method if Browse search gets adapted to the API at some point.

Implementation concerns:

- When you finish an image upload, there's a button which lets you manage metadata for the images you just uploaded. It does this by specifying the image IDs in the request. This may force us to have potentially 1000s of image IDs (each 7 digits) in the URL, which feels... clunky. So for this, I'm thinking of using a specification like 'get images past this upload date / past this ID', rather than 'get images of these IDs'.
- The annotation tool is POST for a couple of reasons right now. First, the current implementation of prev/next navigation requires it, and second, the annotation tool view actually does modify data: it creates an `AnnotationToolAccess` record, and it may apply Alleviate. I think we should also change the annotation tool to GET to make Back/Forward less clunky, but it'll take a couple of extra steps: `AnnotationToolAccess` and Alleviate need to be separated into Ajax requests which run upon loading the annotation tool.

StephenChan:
Done in PR #621.

Note that the annotation tool view is still POST, although the links to Browse Images from that page are now GET. But anyway, converting the annotation tool itself to GET would be a separate issue.

PR #621, opened 2025-05-07
HTTP GET for Browse search

StephenChan:
For issue #437. I figured I'd do this as prep for improving the performance of the Browse pages. However, I'm trying to be careful with this task, since it touches parts like the logic to select images/annotations to delete; so I'm trying to be thorough with testing.

As a result, this PR is snowballing a little more than I hoped it would. It's great work to be getting done, but makes me antsy when there's an important performance task on the queue... oh well, I'm already pretty far in now.

### DIFF
--- a/project/accounts/views.py
+++ b/project/accounts/views.py
@@ -301,7 +301,7 @@ class EmailChangeConfirmView(LoginRequiredMixin, TemplateView):
 def profile_list(request):
     all_profiles = Profile.objects.all().order_by('user__date_joined')
     all_results = [p for p in all_profiles if can_view_profile(request, p)]
-    page_results, _ = paginate(
+    page_results = paginate(
         results=all_results,
         items_per_page=50,
         request_args=request.GET,
--- a/project/annotations/forms.py
+++ b/project/annotations/forms.py
@@ -21,6 +21,7 @@ from accounts.utils import is_robot_user
 from images.models import Metadata, Point
 from labels.models import LocalLabel
 from lib.forms import EnhancedMultiWidget
+from visualization.forms import ResultCountForm
 from .model_utils import AnnotationArea
 from .models import Annotation, AnnotationToolSettings
 
@@ -329,3 +330,16 @@ class ExportAnnotationsForm(Form):
     optional_columns = MultipleChoiceField(
         widget=CheckboxSelectMultiple, choices=optional_columns_choices,
         required=False)
+
+
+class BatchAnnotationDeleteCountForm(ResultCountForm):
+
+    delete_count_mismatch_error_message = (
+        "The number of image results just before deletion"
+        " ({delete_count}) differs from the number shown in the search"
+        " ({expected_delete_count}). So as a safety measure,"
+        " no annotations were deleted."
+        " Make sure there isn't any ongoing activity in this source"
+        " which would change the number of image results. Then,"
+        " redo your search and try again."
+    )
--- a/project/annotations/templates/annotations/annotation_tool.html
+++ b/project/annotations/templates/annotations/annotation_tool.html
@@ -209,12 +209,7 @@
         <br/>
         <br/>
 
-        <form action="{% url 'browse_images' source.id %}" method="post">
-          {% csrf_token %}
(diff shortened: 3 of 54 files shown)


StephenChan:
Okay, finally ready to move on now... general thoughts:

- I forgot that this part of the site can be a bit like whack-a-mole with the corner cases. In fact, some corner cases weren't handled well to begin with, so I had to think about how to handle them elegantly.

- It's good for me to be nervous about batch deletion, but ideally that nervousness would've translated to gradual improvements and added tests, rather than a particularly big round of them like this; oh well.

Some specifics:

- With the search form using GET instead of POST, the request params go into the URL. So it would have created long, ugly URLs with blank params such as `image_name=&aux1=&aux2=&aux3=&aux4=` unless I specifically reworked it to exclude blank params. So, that's just what I did. I also changed a couple of fields so that the default field values would end up with blank params; for example, a sort direction of Ascending produces a blank value (but Descending produces sort_direction=desc). Thus, the fewer fields you change, the fewer params you see in the URL, which is intuitive.

- There used to be an `image_form_type` field in the Browse pages which served a dual purpose: 1) toggling between search filters vs. an image-ID-list filter, and 2) ensuring there's at least one non-blank param submitted in the search form, which allows the delete actions to be submitted. Now the image-ID-list field is treated pretty much the same way as the search fields, so 1 doesn't apply anymore. For 2, there's now a replacement field and value, `search=true`.

- The image_id_list filter is now only used by "the selected images on this page". Coming from an image upload now uses image_id_range, a range defined by two numbers, like `1600_1700`.

  - The advantage of this is that the value doesn't grow arbitrarily as the size of the upload increases, like `1600_1601_1602_1603_...`. (Note: this is more acceptable for the selected images case because that's only up to 20 images; and, I added a check to ensure a manually crafted param doesn't go much higher than that.)

  - The disadvantage of image_id_range: if you happen to have two concurrent uploads in the same source, their min and max image IDs will have overlapping ranges. This means the ranges will be incorrect, since the idea is to have the filter specify exactly the images that were part of the upload operation. However, there isn't a particularly easy alternative other than doing what we did before - accepting an arbitrarily long image_id_list - so I decided this possible incorrectness was acceptable enough.

- Delete images and delete annotations now abort if the image count at the time of deletion doesn't match what the browse page showed. This hopefully adds peace of mind with these batch delete operations, at least for me.
        
  - It remains to be seen whether this feature is more annoying than it's worth in some situations, such as requiring a few retries if automatic image classification keeps changing the number of search results (e.g. if the search is based on classification status).
        
  - Also worth noting that this is a true "time of deletion" count for delete images, since it uses the number returned by the deletion call. But for delete annotations, the count is taken just *before* deleting annotations. I couldn't think of a way to make the latter also a "true" count without greatly sacrificing performance.

- On the unit tests:

  - Existing tests were rather inconsistently structured between the 3 Browse pages and the various 'action' views that use the Browse filters, so in the interest of more careful testing, I took the time to make these test modules more consistent with each other. That sometimes meant function reuse, and sometimes meant copy-paste, trying to strike a balance between less code vs. more explicitness.

  - I also looked at what cases were tested for each Browse page / Browse action, and added tests to make those more consistent.

  - Upload Images now has QUnit tests. Not thorough tests, but at least the most basic upload case is pretty much tested from back to front. This allowed me to have test coverage for the construction of the "proceed to edit metadata" link.

StephenChan:
This PR was meant as groundwork for improving performance of the three Browse pages, but this PR could already give one decent optimization (for some sources) in these pages.

`BaseImageSearchForm.__init__()` used to run DB queries to grab all of the source's distinct values for the aux. metadata fields and about 10 different non-aux. metadata fields. Those queries were meant to populate dropdown options, but said dropdowns for the non-aux. fields were removed a while back, so those particular queries have been doing nothing useful for a while. As part of the cleanups in commit 7d0fd13, the queries for the non-aux. fields have been removed.

Since there are still 5 such queries for the aux. fields, and of course other queries going on in this page too, I wouldn't expect a 3x speedup or anything, but it could be respectable for sources with high image counts or with many unique values on the non-aux. fields.

PR #626, opened 2025-06-18
Multi-column indexes for Annotation model

StephenChan:
Due to the sheer number of Annotations there can be in a source (highest we have seems to be 50k images, 400 points per image), performance of Annotation-query-heavy pages like Browse Patches can be a difficult problem. So I'm explicitly defining the Annotation table's database indexes in an attempt to optimize.

```python
class Annotation(models.Model):
    ...
    class Meta:
        indexes = [
            models.Index(
                fields=['image', 'user'],
                name='annotation_to_img_usr_i'),
            models.Index(
                fields=['image', 'label', 'user'],
                name='annotation_to_img_lbl_usr_i'),
            models.Index(
                fields=['label', 'source'],
                name='annotation_to_lbl_src_i'),
            models.Index(
                fields=['source', 'user', 'robot_version'],
                name='annotation_to_src_usr_rbtv_i'),
            models.Index(
                fields=['source', 'label', 'user'],
                name='annotation_to_src_lbl_usr_i'),
        ]
```

Ideas behind these index definitions:

1) A multi-column index [A,B,C] should be effective for handling queries that filter on just A, on just A + B, or on A + B + C. (Index column order may matter, depending how we want results to be sorted.) If we have such an index [A,B,C], then an [A,B] multi-column index would be redundant, and a single-column index on just A would also be redundant.

2) We seem to almost never filter on just user, or just robot version, so those don't need their own indexes (i.e. indexes where that is the first column).

3) Cutting out indexes we don't use saves us some time when doing inserts. To confirm what we don't use (or almost never use), run `select * from pg_stat_user_indexes;` on the dbshell and look at number of scans.

4) The choices of multi-column indexes are hopefully tailored to our most expensive/frequent Annotation queries.

5) We do not explicitly define indexes for unique constraints (i.e. the OneToOne from Annotation to Point), as those seem guaranteed to be automatically defined, since the constraint isn't allowed to exist without the index.

Now, the annoying part about this PR was that CoralNet has existed since 2011, and it appears that Django wasn't all that tidy about tracking database indexes until [Django](https://docs.djangoproject.com/en/5.2/releases/1.11/#class-based-model-indexes) [1.11](https://github.com/django/django/commit/156e2d59cf92eb252c2f6ef9bb0b65f26c707de2) (released 2017, and part of CoralNet starting 2019). The biggest problem being that the older indexes have different names between environments.

So, to clean up old no-longer-wanted indexes, I couldn't just delete them by name. I had to look up the relevant indexes in the `pg_indexes` table, get their names from there (without accidentally grabbing other indexes), and then delete. I believe this is now working fine in migration 0029, but it just took a little longer to figure out than I'd like, and I was hoping Django would have made this process easier.

The newly added indexes have the index names in both `annotations/models.py` and one of the migration files, so there shouldn't be any future ambiguity on these particular indexes.

Anyway, it remains to be seen how Annotation-query-based pages perform with this change. I've done some testing in staging, but the measure of success will be request durations in production. For example, hopefully it'll further improve the situation of issue #348 (which can be assessed as the time it takes to land on Browse Patches, before even doing a search).


StephenChan:
Getting bitten by (almost certainly) unrelated intermittent tests again. Not really up for dealing with those right now.

PR #631, opened 2025-07-05
Make Image-Metadata relation one-to-one + move it from Metadata to Image

StephenChan:


StephenChan:
Never mind, PR #634 speeds up the migration 10x by using QuerySet.bulk_update().

